PR #3, opened 2026-03-19
Fix CLI sync, Telegram error handling, and memU max_tokens

constkolesnyak:
## Summary
- **CLI sync**: Fix broken `nerve sync` command after `IngestResult` refactor — update field names and remove stale `engine` argument from `build_source_runners`
- **Telegram channel**: Add global error handler, log incoming messages, and wrap error replies in try/except to prevent silent failures
- **memU bridge**: Default `max_tokens` to 4096 when omitted, preventing Anthropic API 400 errors

## Test plan
- [ ] Run `nerve sync` and verify it completes without errors
- [ ] Send a message via Telegram and confirm it's logged and processed
- [ ] Trigger an error in the Telegram handler and verify the error reply is sent (or gracefully logged if reply fails)
- [ ] Verify memU extraction works without explicitly passing `max_tokens`

### DIFF
--- a/nerve/agent/engine.py
+++ b/nerve/agent/engine.py
@@ -997,6 +997,7 @@ class AgentEngine:
         channel: str | None = None,
         model: str | None = None,
         internal: bool = False,
+        images: list[dict[str, Any]] | None = None,
     ) -> str:
         """Run the agent for a user message and return the final text response.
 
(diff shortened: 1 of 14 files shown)


constkolesnyak:
у меня телеграм бот помер. теперь воскрес. надеюсь до утра доживёт

PR #24, opened 2026-04-01
Fix updated source notifications being silently dropped

groeneai:
## Summary

- When a GitHub notification is updated (e.g. new comment on a PR), the source adapter re-fetches it with a newer timestamp
- `INSERT OR IGNORE` silently discards the update because the old row with the same `(source, id)` already exists — the updated notification (with the new comment content) is lost
- **Fix:** Before inserting, `DELETE` the existing row if its timestamp is older. The fresh `INSERT` gets a new `rowid`, making the message appear "unread" for any consumer whose cursor already passed the old rowid
- When timestamps match, the `DELETE` is a no-op — preserving the original dedup behaviour

**Impact:** ~72% of GitHub notifications (reason=`author`) were being silently dropped because the inbox consumer had already advanced past the original rowid, and updated notifications were never re-inserted.

## Test plan

- [x] All 326 existing tests pass
- [x] Verified locally: updated notification gets new rowid, appears as unread to consumer
- [x] Same-timestamp notifications still deduped (INSERT OR IGNORE unchanged)

### DIFF
--- a/nerve/db/sources.py
+++ b/nerve/db/sources.py
@@ -119,6 +119,17 @@ class SourceStore:
         async with self._atomic():
             for r in records:
                 try:
+                    # If this notification was updated (newer timestamp), delete the
+                    # stale row first.  The fresh INSERT then gets a new rowid, which
+                    # makes the message appear "unread" for every consumer whose
+                    # cursor already passed the old rowid.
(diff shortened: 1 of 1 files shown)


groeneai:
Closing — upstream main already landed a more comprehensive fix (checks metadata + content changes, not just timestamp). My change is fully redundant after rebase.

PR #39, opened 2026-04-08
Enrich GitHub PR notifications with review state and inline comments

groeneai:
## Problem

GitHub sets `latest_comment_url` to `null` for PR review submissions. This means when a reviewer submits a `CHANGES_REQUESTED` review with feedback only in inline code comments, the notification delivered to the inbox processor contains **zero comment content**.

**Real-world impact:** @alesapin submitted `CHANGES_REQUESTED` on PR #102033 (ClickHouse) with an inline comment — the inbox processor saw an empty notification, tagged the task `check-comments`, and left it `in_progress` instead of escalating to `pending` with `has-reviewer-feedback`. The review was missed for hours.

### Data flow before this fix

```
GitHub notification:
  reason: "mention"              ← inline comment @mentions us
  latest_comment_url: null       ← GitHub gives us nothing!

Enriched notification delivered to LLM:
  Repository: ClickHouse/ClickHouse
  Title: Fix LOGICAL_ERROR crash in IcebergMetadata...
  Reason: mention
  (no comment, no review state)

LLM: "no visible comment" → tags check-comments → missed
```

### Data flow after this fix

```
Enriched notification delivered to LLM:
  Repository: ClickHouse/ClickHouse
  Title: Fix LOGICAL_ERROR crash in IcebergMetadata...
  Reason: mention

  --- Latest Review (alesapin) ---
  State: CHANGES_REQUESTED

  --- Inline Review Comments ---
  alesapin on src/.../IcebergMetadata.cpp:942:
  @groeneai Let's fix all the places to create table_state_snapshot...

LLM: sees CHANGES_REQUESTED + feedback → escalates immediately
```

## Changes

For `PullRequest` notifications, `_enrich_notification` now also calls `_enrich_pr_reviews` which:

1. **Fetches the latest review** via `pulls/{id}/reviews` — extracts `state` (APPROVED, CHANGES_REQUESTED, COMMENTED), user, body, timestamp
2. **If the review body is empty**, fetches inline comments via `pulls/{id}/reviews/{review_id}/comments` — extracts user, body, path, line for up to 5 comments

Both are rendered into the notification content as `--- Latest Review ---` and `--- Inline Review Comments ---` sections.

**Cost:** +1-2 API calls per PR notification (reviews list + conditionally inline comments). Non-PR notifications are unaffected.

## Testing

- All 326 existing tests pass
- Verified against real GitHub API data from PR #102033

### DIFF
--- a/nerve/sources/github.py
+++ b/nerve/sources/github.py
@@ -275,9 +275,15 @@ class GitHubSource(Source):
         if not isinstance(reviews_data, list) or not reviews_data:
             return
 
+        # Filter out PENDING (draft) reviews — they have submitted_at=null
+        # which would crash max() when comparing str with None.
+        submitted = [r for r in reviews_data if r.get("state") != "PENDING"]
+        if not submitted:
+            return
+
         latest_review = max(
-            reviews_data,
-            key=lambda r: r.get("submitted_at", ""),
+            submitted,
+            key=lambda r: r.get("submitted_at") or "",
         )
         review_state = latest_review.get("state", "")
         result["latest_review"] = {
--- a/nerve/sources/github.py
+++ b/nerve/sources/github.py
@@ -3,8 +3,10 @@
 Cursor semantics: ISO 8601 timestamp of the newest notification's `updated_at`.
 On first run (no cursor), fetches from the last 24 hours.
 
-Each notification is enriched with actual content from the subject (PR/issue)
-and, when available, the latest comment that triggered the notification.
+Each notification is enriched with actual content from the subject (PR/issue),
+the latest comment that triggered the notification, and for PRs the latest
+review state (APPROVED, CHANGES_REQUESTED, etc.) plus any inline review
+comments attached to specific code lines.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


pufit:
### Bug: `max()` will crash on PENDING reviews

```python
latest_review = max(
    reviews_data,
    key=lambda r: r.get("submitted_at", ""),
)
```

GitHub's `/reviews` endpoint returns **all** reviews including `PENDING` ones, which have `"submitted_at": null`. `dict.get("submitted_at", "")` returns `None` (key exists with value null) — not `""`. Comparing `str` with `None` raises `TypeError` in Python 3.

Suggest filtering to submitted reviews first:

```python
submitted = [r for r in reviews_data if r.get("state") != "PENDING"]
if not submitted:
    return
latest_review = max(submitted, key=lambda r: r.get("submitted_at") or "")
```

This also avoids treating a PENDING (draft) review as the "latest" one.

> *Generated by [Nerve](https://github.com/pufit/nerve)*

groeneai:
Fixed — thanks for the catch.

`PENDING` reviews have `submitted_at: null` in the JSON, so `dict.get("submitted_at", "")` returns `None` (key exists with null value), not `""`. The `max()` comparison then crashes with `TypeError: '<' not supported between instances of 'str' and 'NoneType'`.

Fix: filter out `PENDING` reviews before `max()`, and use `or ""` fallback for any remaining null timestamps. Draft reviews shouldn't be treated as the "latest" review anyway.

pufit (approved):

PR #53, opened 2026-04-25
Use lower thinking/effort for cron and hook sessions

constkolesnyak:
## Problem

Every cron run was failing with:

```
API Error: 400 {"error":{"message":"level \"max\" not supported, valid levels: low, medium, high","type":"invalid_request_error"}}
```

Root cause: nerve passed the global `agent.effort=max` and `agent.thinking=max` to **every** session, including cron. Cron sessions run on `cron_model` (Sonnet by default), and under Claude OAuth (subscription) `cli-proxy-api` rejects `level=max` for non-flagship models — only `low`/`medium`/`high` are accepted. With a persistent cron session this poisons the SDK session id, so every subsequent run on the same cron job fails identically until the session is rotated.

This was the second time this hit (first incident: April 21, 2026, same error in `inbox-processor`).

## Fix

- Add `agent.cron_thinking` and `agent.cron_effort` fields (default `high`) — separate knobs for cron/hook sessions.
- Extract `_select_thinking_effort(agent_config, source)` helper that returns `(thinking, effort)` based on session source — `cron` and `hook` get the cron overrides, everything else keeps the main settings.
- Wire it into `_build_options(...)` so interactive sessions still get the full thinking budget on the flagship model, while cron jobs stay within what their proxy/provider accepts.

## Test plan

- [x] `tests/test_engine_options.py` — 13 new tests covering source routing, default values, custom overrides, and back-compat for configs that don't define the new fields
- [x] Full suite passes: `344 passed, 2 skipped`


constkolesnyak:
Wrong target repo, recreating against constkolesnyak/nerve

PR #56, opened 2026-04-27
fix(channels): deliver send_file via Telegram and harden cross-channel dispatch

neomnezia:
## Summary

`mcp__nerve__send_file` was leaving phantom `[mcp__nerve__send_file]` headers in Telegram and only persisting a DB record for the web frontend — files never reached the chat. This PR makes the tool actually deliver via the bound channel and closes three P1 cross-channel leakage paths.

### What changed

- New `SEND_FILES` channel capability + `BaseChannel.send_file` interface.
- Telegram `send_document` impl with path/exists checks. No client-side size cap — the Telegram API enforces per-deployment limits (50 MiB on `api.telegram.org`, 2 GiB on self-hosted Bot API), and the surrounding try/except already handles the rejection.
- Web `send_file` no-ops since `SendFileBlock` already handles UI delivery.

### Cross-channel hardening

- `AgentEngine` tracks active channel per session (`_active_channel` set in `run()`, cleared on exit) and exposes `get_active_channel(session_id)`.
- `ChannelRouter.send_file(session_id, file_path, channel=...)` requires an explicit channel. Cached `_message_context` target is reused only when its channel matches the requested one — never to **pick** the destination channel. `channel=None` returns False unconditionally, so cron/planner/notifications/web routes that don't thread a channel through `engine.run()` can no longer leak files to a stale Telegram chat from a prior inbound message.
- `_send_file_impl` reads `engine.get_active_channel` and forwards.
- Path-aware workspace containment check (`Path.relative_to` / `ValueError`) replaces the bypassable `str.startswith` guard — closes a sibling-prefix exfiltration vector exposed by real Telegram delivery (e.g. workspace `/srv/ws` previously accepted `/srv/ws-evil/secret.txt`).

### Files

- `nerve/channels/base.py` — `Capability.SEND_FILES`, `BaseChannel.send_file` abstract.
- `nerve/channels/telegram.py` — `send_file` via `bot.send_document`, capability advertised.
- `nerve/channels/web.py` — no-op (UI already covered).
- `nerve/channels/router.py` — explicit-channel dispatch, cached-context match guard.
- `nerve/agent/engine.py` — `_active_channel` map, `get_active_channel()`, run()/exit lifecycle.
- `nerve/agent/tools.py` — `_send_file_impl` resolves channel via engine, path containment check.
- `tests/test_send_file.py` — 28 new tests (router dispatch, Telegram channel, tool impl, engine accessor, leakage + sibling-prefix bypass scenarios).

## Test plan

- [x] `pytest tests/test_send_file.py -v` → 28/28 pass
- [x] Full suite: 394 passed, 2 skipped
- [x] Smoke-tested end-to-end on Telegram: agent-invoked `send_file` on a workspace file delivers the document to the chat
- [x] Secret/PII scan on diff — clean

### DIFF
--- a/nerve/agent/engine.py
+++ b/nerve/agent/engine.py
@@ -240,9 +240,17 @@ class AgentEngine:
         self._skill_manager: SkillManager | None = None
         self._memorize_lock = asyncio.Lock()
         self._session_locks: dict[str, asyncio.Lock] = {}
+        # Per-session active channel — set on run() entry, cleared on exit.
+        # Read by session-scoped tools (send_file) to avoid dispatching via
+        # stale router context from a prior inbound channel.
+        self._active_channel: dict[str, str] = {}
         self._router = None  # ChannelRouter — lazy-initialized via .router property
         self._mcp_servers_cache = list(config.mcp_servers)  # hot-reloadable
         self._claude_code_plugins: list[dict[str, str]] = []  # plugin dirs
+
+    def get_active_channel(self, session_id: str) -> str | None:
+        """Return the channel name currently driving ``session_id`` (or None)."""
+        return self._active_channel.get(session_id)
 
     async def initialize(self) -> None:
         """Initialize the agent engine — set up tools and main session."""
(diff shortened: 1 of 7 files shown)


pufit (approved):
Thank you for your contribution! This is a great feature!

PR #64, opened 2026-05-05
agent/streaming: ship synthetic done when a turn ends without one

alex-fedotyev:
## Summary

Chats appear stuck mid-step: the session card drops out of the sidebar's "Running" group into the time-grouped list, but the chat detail keeps showing the "thinking..." spinner because isStreaming was never reset.

The previous c439473 fix (tool schemas aborting agent turns) closed one underlying cause. Three remaining cases also leave `_run_inner` without emitting a terminal event:

1. **Post-stream exception.** After `receive_response` yields `ResultMessage` the loop exits cleanly, but the ~80 lines after that contain unguarded post-stream work (`add_message`, `mark_active`, `update_session_metadata`, `record_turn_usage`). If any of those raise, the exception propagates up to `engine.run()` before `broadcast_done` ever fires.
2. **Hung CLI cancelled externally.** `claude_agent_sdk`'s `receive_response` documents "If no ResultMessage is received, the iterator continues indefinitely." When the CLI wedges and the asyncio task is cancelled, the CancelledError handler sends `stopped` on the normal path, but the WS message can be lost mid-reconnect.
3. **Lost WebSocket message during reconnect.** The frontend handler had no symmetry: `handleSessionRunning` enters streaming on `is_running:true` but never exits on `is_running:false`.

I confirmed case 1 against a live session in our DB: `last_activity_at` stuck at `connected_at` (04:13:13) but `updated_at` advanced to 04:30:29, consistent with `mark_active` running but a later DB write failing before `broadcast_done`.

## What changed

**Server.** `StreamBroadcaster` gains an `_open_turns` set with `mark_turn_open` / `is_turn_open` / `clear_turn_open`. `broadcast()` auto-clears the flag when a terminal event ships (done/stopped/error). `engine.run()` calls `mark_turn_open` before `_run_inner`; its `finally` checks `is_turn_open` and ships a synthetic `broadcast_done` if still set. Single backstop covers all three cases at the source.

**Frontend.** `handleSessionRunning` gets a symmetric exit branch. When the active session reports `is_running:false` while `isStreaming` is true, it commits any in-flight streaming blocks to messages, clears `streamingBlocks`, and resets `agentStatus` to idle. Belt-and-suspenders for the case where even the synthetic `done` is missed (e.g. WS reconnect drops the done event but redelivers `session_running:false`).

## Test plan

- [x] `tests/test_streaming.py::TestOpenTurnTracking`: 5 new tests covering each terminal type clearing the flag, non-terminal types not clearing, explicit `clear_turn_open`, multi-session isolation, and the `broadcast_done` helper closing the turn
- [x] `tests/test_streaming.py` + `tests/test_sessions.py` + `tests/test_engine.py`: 76/76 pass
- [x] `web/` tsc -b clean
- [ ] Smoke: open a chat, run a turn that exercises post-stream cleanup, confirm the spinner clears

### DIFF
--- a/nerve/agent/engine.py
+++ b/nerve/agent/engine.py
@@ -1427,6 +1427,11 @@ class AgentEngine:
                 self.sessions.mark_running(session_id)
                 if channel is not None:
                     self._active_channel[session_id] = channel
+                # Mark the turn as in flight so the finally below can
+                # detect "ended without sending done/stopped/error" and
+                # ship a synthetic done.  Clearing happens automatically
+                # when a terminal event is broadcast.
+                broadcaster.mark_turn_open(session_id)
                 # Notify all connected clients that this session started running
                 await broadcaster.broadcast("__global__", {
                     "type": "session_running",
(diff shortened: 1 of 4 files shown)


alex-fedotyev:

pufit (approved):

PR #65, opened 2026-05-05
agent/engine: guard against stale sdk_session_id when .jsonl is missing

alex-fedotyev:
## What and why

When a Nerve container restarts without a persisted `/root/.claude` bind mount, the Claude Code `.jsonl` conversation files are wiped. The Nerve DB still holds the old `sdk_session_id`, so the next turn invokes the CLI with `--resume <stale-id>` and the CLI exits 1 immediately: "No conversation found with session ID." The session hangs mid-turn.

This PR has two fixes:

**1. Engine guard** (`nerve/agent/engine.py`)

Added `_sdk_resume_file_exists(sdk_session_id)` which checks whether `~/.claude/projects/<encoded-cwd>/<sdk_session_id>.jsonl` is on disk before passing `--resume` to the CLI. If the file is missing, the stale `sdk_session_id` is cleared in the DB and the turn starts a fresh conversation instead of crashing. On any unexpected error the method returns `True` (fail-open) so genuine bugs still surface via the CLI.

Fork-source sessions are exempt: a fork's resume target is the source session's .jsonl, which lives on its own row.

**2. Bootstrap mount** (`nerve/bootstrap.py`)

Added `~/.nerve/claude:/root/.claude` to the generated `docker-compose` so conversation history survives `docker compose down/up`. The mount lives under `~/.nerve` to keep the agent's CLI state isolated from the host user's personal `~/.claude` (macOS stores OAuth tokens there).

The entrypoint now runs `mkdir -p /root/.claude && chmod 700 /root/.claude` on startup so the bind-mounted directory is always accessible before any tool touches it.

## Tests

Five new cases in `tests/test_engine.py`:

- file present → `True`
- file absent → `False`
- `os.path.isfile` raises → `True` (fail-open)
- workspace `/` encoded to `-` in the projects subdir path
- path always ends with `<session_id>.jsonl`

All 23 tests pass.

## Related

Companion to #64 (synthetic done events for stuck sessions). That PR closes the streaming-state bug; this one closes the resume-target bug.

### DIFF
--- a/nerve/agent/engine.py
+++ b/nerve/agent/engine.py
@@ -1092,6 +1092,39 @@ class AgentEngine:
             return True
         return process.returncode is not None
 
+    def _sdk_resume_file_exists(self, sdk_session_id: str) -> bool:
+        """Check whether Claude Code still has the conversation .jsonl
+        for the given SDK session ID on this filesystem.
+
+        The CLI stores history at::
+
+            ~/.claude/projects/<encoded-cwd>/<sdk_session_id>.jsonl
+
+        where <encoded-cwd> is the absolute cwd path with every '/'
+        replaced by '-'.  If the file is gone (typically because the
+        container's /root/.claude was not bind-mounted and got wiped on
+        restart), passing --resume to the CLI fails with exit 1.
+
+        Best-effort check: any unexpected error returns True so we still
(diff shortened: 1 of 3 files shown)


alex-fedotyev:

pufit (approved):

PR #129, opened 2026-06-20
Use lower thinking/effort for cron and hook sessions under OAuth

constkolesnyak:
## Problem

Every cron run on Claude OAuth (subscription) fails with:

```
API Error: 400 level "max" not supported, valid levels: low, medium, high
```

Root cause: the global `agent.thinking=max` / `agent.effort=max` defaults are applied to **every** session, including cron and hook sessions that run on `cron_model` (Sonnet by default). Claude OAuth caps non-flagship models at `high` and rejects `max`. With a persistent cron session this also poisons every subsequent run on the same `sdk_session_id`.

The main interactive model (Opus) accepts `max`, so just lowering the global default would degrade the interactive experience to fix cron.

## Fix

Add dedicated `agent.cron_thinking` and `agent.cron_effort` settings (default `high`) and a small `_select_thinking_effort(agent_config, source)` helper that returns the right pair for the session source:

- `source in {"cron", "hook"}` → `cron_thinking` / `cron_effort`
- everything else → `thinking` / `effort` (unchanged)

Defaults match what works under OAuth on Sonnet (`high`). Users with a provider that accepts `max` for Sonnet can lift the defaults.

Composes cleanly with the existing model-aware `_parse_thinking_config(value, model)` / `_effective_effort(value, model)` helpers — `_select_thinking_effort` only picks the source-appropriate input, the model-aware step still gates per-model legality.

## Tests

`tests/test_engine_options.py` (new, 13 tests, all passing):

- `_select_thinking_effort` returns cron_* for `cron` and `hook`, defaults for everything else
- explicit override + None handling
- defaults of `high` for cron_thinking / cron_effort

Full suite: `1164 passed, 2 skipped` (excluding pre-existing `tests/test_cron.py::TestMaybeRotateContext` rotate_at failures unrelated to this change — present on `main` too).

## Files

- `nerve/config.py` — add `cron_thinking` / `cron_effort` to `AgentConfig` with `"high"` defaults
- `nerve/agent/engine.py` — `_select_thinking_effort` helper, used at session-start path
- `nerve/bootstrap.py` — surface in bootstrap defaults
- `config.example.yaml`, `docs/config.md` — document the two new keys
- `tests/test_engine_options.py` — new

## History

Previously opened as [#53](https://github.com/ClickHouse/nerve/pull/53), closed in error against the wrong target. This is the recreated version against upstream, rebased onto current `main` with the model-aware `_parse_thinking_config` / `_effective_effort` API.

> *Generated with [Claude Code](https://claude.com/claude-code)*

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -20,6 +20,15 @@ agent:
   title_model: claude-haiku-4-5-20251001  # Session title generation
   max_turns: 50                  # Max agentic turns per request
   max_concurrent: 4              # Max concurrent agent sessions
+  thinking: max                  # Thinking budget for the main model
+  effort: max                    # Reasoning effort for the main model
+  # Cron / hook sessions use the cron_model (Sonnet). Under Claude OAuth
+  # (subscription) Sonnet caps thinking budget and rejects "max" with
+  #   level "max" not supported, valid levels: low, medium, high
+  # Keep these at "high" unless your provider accepts "max" for the
+  # cron model.
+  cron_thinking: high
+  cron_effort: high
   # First-prompt rewrite — the web UI can refine the opening message of a
   # new chat with a fast model, preview it, and send only after approval.
   prompt_rewrite:
--- a/docs/config.md
+++ b/docs/config.md
@@ -40,6 +40,10 @@ from any working directory:
 | `agent.cron_model` | string | `claude-sonnet-4-6` | Model for cron jobs (cheaper) |
 | `agent.max_turns` | int | `50` | Max agentic turns per request |
 | `agent.max_concurrent` | int | `4` | Max concurrent agent sessions |
+| `agent.thinking` | string | `max` | Thinking budget for the main model: `max` / `high` / `medium` / `low` / `disabled` / `adaptive` / explicit token count |
+| `agent.effort` | string | `max` | Reasoning effort for the main model: `max` / `high` / `medium` / `low` |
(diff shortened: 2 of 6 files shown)


config.example.yaml
```diff
@@ -20,6 +20,15 @@ agent:
...
   max_turns: 50                  # Max agentic turns per request
   max_concurrent: 4              # Max concurrent agent sessions
+  thinking: max                  # Thinking budget for the main model
+  effort: max                    # Reasoning effort for the main model
+  # Cron / hook sessions use the cron_model (Sonnet). Under Claude OAuth
+  # (subscription) Sonnet caps thinking budget and rejects "max" with
+  #   level "max" not supported, valid levels: low, medium, high
+  # Keep these at "high" unless your provider accepts "max" for the
+  # cron model.
+  cron_thinking: high
```
pufit, line 30: This will change the behavior for API users. We should keep it `max` by default and switch to `high` when OAuth subscription is used.

pufit (changes requested):

constkolesnyak:
Superseded by #153 — does exactly what you asked: keeps `max` by default and switches to `high` only when OAuth subscription is in use (detected via `config.proxy.enabled`). The `agent.cron_thinking` / `agent.cron_effort` config knobs introduced here are removed in #153 in favor of the implicit OAuth gate.

PR #135, opened 2026-06-20
feat(sidebar): replace search field + header with Search/New chat pills

constkolesnyak:
## Summary

Replaces the static "Conversations" header + always-visible search input with two rounded pills: **Search sessions** (left) and **New chat** (right). The search input mounts on hover/focus, fades in over 200ms, and unmounts again when nothing is interacting with it. Reclaims sidebar vertical space without losing functionality.

## Behaviour

- **Search pill** is the hover-zone trigger. On hover (or when a query is non-empty), the input mounts and fades in; on leave + empty query, it fades out and unmounts. Never lingers in the DOM when unused.
- **New chat pill** sits on the right. While the search input is open, the input overlays the pill (z-20) so the create button is hidden behind the field — no visual clutter.
- Both controls share `rounded-full` + subtle border styling so they read clearly as buttons.

## Why pills, not a static field

The previous design committed a full row of vertical space to a permanently-visible search field even though searching is intermittent. The pill stays compact when idle and only grows when you actually reach for it. The 200ms fade is short enough not to feel laggy on a deliberate hover; the unmount-on-leave keeps the DOM small.

Subjective UI — happy to drop the redesign or rework if it doesn't fit upstream taste.

## Tests

`npm run build` clean (`✓ built in 8.45s`). Web-only change, no backend touch, no new tests.

Verified by hand: hover → input appears → focus retained while typing → blur + leave → input unmounts.

## Files

- `web/src/components/Chat/SessionSidebar.tsx` — pills + mount/fade state (`searchHovered` / `searchFocused` / `searchMounted` / `searchVisible`) (+90 / -31)

## Note

The original commit (`7fa6505`) was followed by three sequels (`c4617fc`, `e058d11`, `3cd0e1f`) that wired the pill into the global Cmd+K shortcut and added a `requestSearchFocus()` store action. Those are intertwined with the keyboard-shortcuts work and are not part of this PR — they belong in [#133](https://github.com/ClickHouse/nerve/pull/133) (or a follow-up after both this PR and #133 land).

> *Generated with [Claude Code](https://claude.com/claude-code)*

### DIFF
--- a/web/src/components/Chat/SessionSidebar.tsx
+++ b/web/src/components/Chat/SessionSidebar.tsx
@@ -42,12 +42,50 @@ export function SessionSidebar({ sessions, activeSession, agentStatus, onSelect,
 }) {
   const [systemExpanded, setSystemExpanded] = useState(false);
   const [localQuery, setLocalQuery] = useState('');
+  const [searchHovered, setSearchHovered] = useState(false);
+  const [searchFocused, setSearchFocused] = useState(false);
+  const [searchMounted, setSearchMounted] = useState(false);
+  const [searchVisible, setSearchVisible] = useState(false);
+  const closeTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const debounceRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const inputRef = useRef<HTMLInputElement>(null);
 
   const { searchResults, searchLoading, searchSessions, clearSearch, renameSession, toggleStar } = useChatStore();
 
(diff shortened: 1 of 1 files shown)


constkolesnyak:
Consolidated into #133 — the pills change and the keyboard-shortcut work are intertwined enough (the Cmd+K fixes need both) that splitting them produced cross-PR dependencies that wouldn't merge cleanly. #133 now contains the original commits from this PR plus the shortcut feature and the two follow-up Cmd+K fixes, in chronological order.

PR #147, opened 2026-06-24
Add local Ollama model selection to the web composer

Wachynaky:
Expose locally-installed Ollama models as selectable chat models in the composer's model picker. The Claude Agent SDK only speaks the Anthropic Messages API, so Ollama (OpenAI-compatible) is reached through the bundled CLIProxyAPI proxy, registered as an openai-compatibility upstream — this requires proxy.enabled.

Backend:
- OllamaConfig + ollama_routable gate (Ollama enabled AND proxy enabled)
- nerve/ollama.py: best-effort model discovery via Ollama GET /api/tags
- proxy/service.py: register discovered models as a proxy upstream
- GET /api/models route for the picker
- engine: thread per-session model, recreate the SDK client on a mid-session model switch, and suppress Anthropic-only knobs (extended thinking, effort, context-1m beta) for non-Claude models
- server: pass the WS per-message model through to run()
- startup warning when ollama.enabled but proxy.enabled is false

Frontend:
- api.getModels() + optional model arg on ws.sendMessage
- chatStore holds available/selected/default model (persisted to localStorage)
- ChatInput renders a model picker, shown only when more than one model is offered

config.example.yaml: document the proxy and ollama blocks.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -11,6 +11,25 @@ timezone: America/New_York
 #   aws_region: us-east-1      # AWS region for Bedrock
 #   # aws_profile: ""          # Optional: AWS SSO profile name
 
+# Local proxy (CLIProxyAPI) — optional. Routes Anthropic API calls through
+# Claude Code OAuth, and is the Anthropic↔OpenAI translation layer that local
+# Ollama models are reached through. Required for the ollama block below.
+# proxy:
+#   enabled: false
+#   port: 8317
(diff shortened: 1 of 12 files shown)


Wachynaky:
Sorry, I guess this isn’t the right way or the right code for this feature, but I was doing it locally and in the end the Nerve agent uploaded it itself, with my help running the Git commands!

pufit (approved):

PR #153, opened 2026-06-25
Cron/hook thinking-effort: downgrade only under OAuth proxy

constkolesnyak:
## Problem

Every cron run on Claude OAuth (subscription) fails with:

```
API Error: 400 level "max" not supported, valid levels: low, medium, high
```

Root cause: the global `agent.thinking=max` / `agent.effort=max` defaults are applied to **every** session, including cron and hook sessions that run on `cron_model` (Sonnet by default). Claude OAuth caps non-flagship models at `high` and rejects `max`. With a persistent cron session this also poisons every subsequent run on the same `sdk_session_id`.

The main interactive model (Opus) accepts `max`, so just lowering the global default would degrade the interactive experience to fix cron. API users (direct `ANTHROPIC_API_KEY`) accept `max` on every model and shouldn't be downgraded at all.

## Fix

Detect OAuth via `config.proxy.enabled` — the local cli-proxy-api that wraps Claude Code OAuth credentials is the single source of truth for "running under OAuth subscription" in nerve. Cap `thinking` and `effort` to `high` only when:

- `source in ("cron", "hook")`, AND
- `config.proxy.enabled` is True

API users (no proxy) keep their configured value untouched for every source, and interactive sessions keep it under OAuth too — only the narrow OAuth+cron path is downgraded.

Implementation lives in a small `_select_thinking_effort(config, source)` helper at the session-start path in `nerve/agent/engine.py`. It composes cleanly with the existing model-aware `_effective_effort(value, model)` (which already caps Sonnet's effort at `high`) — this layer also handles the `thinking` side (which the CLI otherwise rejects as a string-level `max`) and makes the OAuth trigger explicit.

No new config knobs.

## Tests

`tests/test_engine_options.py` (new, 21 tests, all passing):

- OAuth ON + `cron` / `hook` + `max,max` → `("high","high")` (the exact bug being fixed).
- OAuth ON + `cron` / `hook` + lower values → unchanged (cap, not force).
- OAuth ON + `cron` / `hook` + mixed → each knob capped independently.
- OAuth ON + interactive sources (`web` / `telegram` / `discord` / `api` / `""`) + `max,max` → unchanged (Opus accepts max under OAuth).
- OAuth OFF + every source + `max,max` → unchanged (Artem's exact requirement).
- OAuth OFF + cron + lower values → unchanged.
- `AgentConfig.from_dict({"cron_thinking": "high", "cron_effort": "high"})` silently ignores those keys (backwards-compat for anyone whose config had them).

## Files

- `nerve/agent/engine.py` — `_select_thinking_effort` helper + single call site update
- `tests/test_engine_options.py` — new

## History

Supersedes [#129](https://github.com/ClickHouse/nerve/pull/129) — addresses [@pufit](https://github.com/pufit)'s review feedback:

> *This will change the behavior for API users. We should keep it `max` by default and switch to `high` when OAuth subscription is used.*

The previous PR introduced two config knobs (`agent.cron_thinking`, `agent.cron_effort`) defaulting to `high` and applied them unconditionally — silently downgrading API users. This version detects OAuth automatically via `config.proxy.enabled` with no new config surface.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -11,25 +11,6 @@ timezone: America/New_York
 #   aws_region: us-east-1      # AWS region for Bedrock
 #   # aws_profile: ""          # Optional: AWS SSO profile name
 
-# Local proxy (CLIProxyAPI) — optional. Routes Anthropic API calls through
-# Claude Code OAuth, and is the Anthropic↔OpenAI translation layer that local
-# Ollama models are reached through. Required for the ollama block below.
-# proxy:
-#   enabled: false
-#   port: 8317
-#   host: 127.0.0.1
-
-# Local Ollama — expose locally-installed Ollama models as selectable chat
-# models in the web composer's model picker. Ollama speaks an OpenAI-compatible
-# API, so this requires proxy.enabled: true (the proxy translates the
-# Anthropic requests the SDK emits). Models are auto-discovered at runtime
-# from the running Ollama server (GET /api/tags) — whatever you've pulled
-# locally shows up automatically, no need to list models here.
-# ollama:
-#   enabled: false
-#   host: 127.0.0.1
-#   port: 11434
-
 # Agent
 agent:
   model: claude-opus-4-8         # Primary model for conversations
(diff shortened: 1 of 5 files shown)


constkolesnyak:
Closing as superseded by upstream's own work.

Two changes since this PR was opened solve the problem it was written for (`400 level "max" not supported` on cron turns) more cleanly than proxy-detection did:

- `ClaudeBackend._MODEL_EFFORT_LEVELS` / `_effective_effort()` now cap effort per-model, so an unsupported level is clamped rather than rejected — regardless of whether the call goes through the OAuth proxy or a direct API key.
- `agent.cron_effort` (with `_base_effort_for_source()`) gives cron/hook turns their own effort setting, which is the knob this PR was really reaching for.

My fork has dropped the patch accordingly — `nerve/agent/backends/claude.py` is now byte-identical to `clickhouse/main`. Nothing here is worth rebasing. Thanks!

PR #170, opened 2026-07-06
Discover the account's Anthropic models to give vhoice in composer model picker

GrigoryPervakov:
## Summary
The Claude backend offers a single model in the composer's model picker — the configured `agent.model`. This PR discovers the Anthropic chat models available to the account and offers the newest one of each family (opus / sonnet / haiku / …), so switching a turn to a stronger or cheaper model needs no configuration.

- `nerve/anthropic.py` — discovery via `GET /v1/models` with `anthropic_api_key`, mirroring the Ollama discovery module: stdlib `urllib`, bounded timeout, never raises — any failure yields an empty list. `latest_per_family()` keeps the first (newest, in API order) model of each family; the family is the first alphabetic token after the `claude-` prefix, which also covers legacy `claude-3-5-sonnet-…` ids.
- `nerve/gateway/routes/models.py` — the discovered list feeds the existing per-backend `models[]` contract (claude backend option + flat picker list), with the default model always first. Results are cached (15 min on success, 60 s retry on failure) since the call leaves the box and `/api/models` fires on every composer mount; discovery runs via `asyncio.to_thread` to keep the event loop free.
- Keyless (OAuth/Bedrock) or offline setups degrade to offering only the default model.
- No frontend changes — the picker already renders per-backend model lists, exactly as it does for Codex preflight and Ollama discovery.

## Test plan
- [ ] With `anthropic_api_key` set: the Claude-backend picker offers the family-latest models (currently sonnet-5, fable-5, opus-4-8, haiku-4-5)
- [ ] Selecting a discovered model routes the turn to it
- [ ] No API key: only the default model → the picker stays hidden
- [ ] Offline: `/api/models` stays fast after the first miss (60 s retry cache) and degrades to the default model
- [ ] `pytest tests/test_anthropic.py`

### DIFF
--- a/nerve/anthropic.py
+++ b/nerve/anthropic.py
@@ -1,0 +1,88 @@
+"""Anthropic API integration helpers.
+
+This module only handles model *discovery* — querying which chat models
+the configured Anthropic API key can use (``GET /v1/models``) so the web
+composer's model picker can offer them without a hand-maintained list.
+
+Discovery is best-effort and never raises: without an API key (OAuth or
+Bedrock setups) or with the API unreachable, callers get an empty list
+and the picker falls back to the configured models.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import urllib.error
+import urllib.request
+
+logger = logging.getLogger(__name__)
+
+_MODELS_URL = "https://api.anthropic.com/v1/models?limit=100"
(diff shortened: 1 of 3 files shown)


pufit:
unfortunately outdated

PR #183, opened 2026-07-18
Fix session stickiness: advance last_activity_at on turns and explicit switches

serxa:
Two related fixes to session-routing stickiness. `last_activity_at` drives `_is_within_sticky_period`, which decides whether an inbound message resumes the channel's mapped session or mints a fresh one.

## 1. `last_activity_at` frozen at connect time
`mark_active()` set `connected_at` and `last_activity_at` to the same value, and its callers pass the session's **original** `connected_at` when resuming an SDK client — so `last_activity_at` was pinned to session start and never advanced across turns. Any session older than `sessions.sticky_period_minutes` was then wrongly rotated onto a fresh session on the next message (an unexpected context reset on Telegram/web). Fix: `connected_at` keeps the caller value (or now); `last_activity_at` **always** advances to now.

## 2. Explicit switches ignored past the sticky period
`set_active_session()` only rewrote the channel→session mapping. Channels without a per-message session id (Telegram) resolve the target via `get_active_session`, whose sticky-period check rejects a session idle longer than the window and mints a fresh one — so selecting a long-idle session and then sending a message **silently routed into a NEW session** instead of the chosen one. Fix: an explicit switch marks the chosen session freshly active (this also bumps `updated_at`, resetting the idle-archive clock), so the deliberate choice is honoured.

## Tests
Regression tests for both: `mark_active` advances `last_activity_at` despite an old `connected_at`; a resumed long-lived session stays sticky after a turn; and an explicit switch to a long-idle session routes the next message there instead of forking. Both fail on `main`, pass with the fix. Full suite: **1550 passed, 2 skipped**. Backend-only.

### DIFF
--- a/nerve/agent/sessions.py
+++ b/nerve/agent/sessions.py
@@ -172,11 +172,21 @@ class SessionManager:
         sdk_session_id: str | None = None,
         connected_at: str | None = None,
     ) -> None:
-        """Mark session as active with an SDK client."""
-        now = connected_at or datetime.now(timezone.utc).isoformat()
+        """Mark session as active with an SDK client.
+
+        ``connected_at`` is the (stable) time the SDK client connected; callers
+        pass the original value when resuming an existing client so it is
+        preserved across turns. ``last_activity_at`` must instead always advance
+        to the real current time, because it drives channel stickiness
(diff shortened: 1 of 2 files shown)


serxa:
This is ready to be reviewed

pufit (approved):

PR #184, opened 2026-07-18
Telegram /sessions: inline-keyboard switch + catch-up tail (expandable, timestamped)

serxa:
Makes `/sessions` a native, mobile-friendly way to see, switch, and catch up on sessions — replacing the old plain-text id list.

## 1. Inline-keyboard session switching
- one **tap-to-switch** button per session (current marked ✓); the id rides in `callback_data` (`sess:<id>`) — no id copy-paste on mobile;
- a **➕ New session** button that starts a fresh session **without stopping the current one** (unlike `/new`). Output is bound per-session (`engine._active_channel[session_id]`), so a session you switch away from keeps running and still delivers to the chat;
- **empty (0-message) sessions are hidden** — not useful switch targets; the list stops counting once the keyboard is full.

## 2. Catch-up history tail on switch
Switching to a session — especially a background one you weren't following — shows its recent history so you can catch up:
- **native chat order** — oldest at the top, most recent at the bottom;
- each message body is an **expandable blockquote** (Bot API 7.0+): long messages collapse to a few lines and **expand in place on tap** — no hard truncation;
- **newest-priority budget**: the most recent message (usually the important one) gets the largest share and is never starved; older messages fill the remainder newest-first. The budget is measured on the assembled, HTML-escaped string, so entity expansion can't push the card past Telegram's 4096-char limit;
- each line carries a **timestamp in the user's timezone** (`config.timezone`), matching the Telegram client;
- **⬆️ Load more** fetches more older history (informational; never changes routing); **🗂 Sessions** returns to the list. No LLM recap by design — ask inside the session.

> Note: honouring the switch for a long-idle session (so the next message actually routes there) is the companion core fix in the session-stickiness PR.

## Reuse (no new infrastructure)
`InlineKeyboardButton`/`Markup` (as in `notifications/service.py`); the single existing `CallbackQueryHandler`, extended with `sess:`/`sesstail:` branches; router helpers `list_sessions`/`switch_session`/`create_session`/`get_last_session` + thin new `get_session`/`get_session_messages`/`count_session_messages` over `engine.db`.

`/session <id>` text command kept for back-compat. Only interactive (telegram/web) sessions are listed; `callback_data` guarded against the 64-byte cap.

## Tests
`tests/test_telegram_sessions.py` (19 tests): switch-list (current-marked, ids in callback_data, New button, empty-session exclusion, keyboard cap) and tail (native order, timezone conversion, expandable blockquote + HTML escaping, newest-priority budget, load-more windowing, 4096-limit safety). Full suite: **1565 passed, 2 skipped**. Backend-only.

### DIFF
--- a/nerve/channels/router.py
+++ b/nerve/channels/router.py
@@ -430,6 +430,20 @@ class ChannelRouter:
         """List sessions, most recently updated first."""
         return await self.engine.sessions.list_sessions(limit=limit)
 
+    async def get_session(self, session_id: str) -> dict[str, Any] | None:
+        """Fetch a session row (title/status/…), or None if it is gone."""
+        return await self.engine.db.get_session(session_id)
+
+    async def get_session_messages(
+        self, session_id: str, limit: int,
+    ) -> list[dict[str, Any]]:
+        """Recent messages for a session, oldest→newest (native chat order)."""
+        return await self.engine.db.get_messages(session_id, limit=limit)
+
+    async def count_session_messages(self, session_id: str) -> int:
+        """Total message count for a session (to know if more history exists)."""
+        return await self.engine.db.count_messages(session_id)
+
     # ------------------------------------------------------------------ #
     #  Outbound: engine → channel (cron delivery, etc.)                    #
     # ------------------------------------------------------------------ #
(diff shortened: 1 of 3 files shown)


serxa:
This is ready to review and merge now

pufit (approved):

PR #185, opened 2026-07-18
Make systemd restart bounded so a self-restart cannot hang nerve down

constkolesnyak:
A bare SIGTERM in the systemd restart branch can hang the service indefinitely.

uvicorn's graceful shutdown waits for in-flight connections to drain. When `nerve restart` is triggered from *inside* a live session — the Telegram bot, or an agent turn editing nerve's own source — that session's stream keeps a connection open, so the drain never completes. systemd's TimeoutStopSec→SIGKILL only escalates for a systemd-initiated stop, not for a process shutting itself down after a signal, so nothing forces the issue.

The result is the worst kind of outage: the gateway stops listening, `Application shutdown complete` never arrives, the process stays alive so `Restart=always` never fires, and `systemctl` keeps reporting the unit **active** while it serves nothing. The scheduler goes on ticking in the logs the whole time, which makes it look healthy from every angle except the port.

## Fix

Spawn a detached helper (own session, survives the caller's death) that sends SIGTERM, waits up to 15s for a clean drain, then SIGKILLs. Once the old PID is gone, `Restart=always` brings up a fresh instance. Bounded, and no privileges or `systemctl` access needed.

Verified on a systemd deployment that reproduced the hang: restart now completes in ~3s with a clean `Application shutdown complete`. 29 CLI tests pass.

### DIFF
--- a/nerve/cli.py
+++ b/nerve/cli.py
@@ -383,15 +383,61 @@ def restart(ctx: click.Context) -> None:
         ctx.exit(rc)
         return
 
-    # Systemd mode (Restart=always): just kill the process — systemd
-    # will bring it back automatically.
+    # Systemd mode (Restart=always): kill the old process and let systemd
+    # respawn it. A bare SIGTERM is NOT enough: uvicorn's graceful shutdown
+    # waits for in-flight connections to drain, and when the restart is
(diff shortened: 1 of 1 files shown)


constkolesnyak:

PR #186, opened 2026-07-18
Fail loudly when two migrations share a version number

constkolesnyak:
`run_migrations` skips anything at or below the recorded schema version, so if two files claim the same number, one silently wins and the other **never runs** — on every database, forever, with no error.

It surfaces much later as a missing column or index, and which of the two is missing depends on filesystem iteration order, so it reproduces on some machines and not others. The traceback you eventually get points at the query, not at the migration that never ran.

The collision is easy to introduce: two branches each add the next migration in parallel, both merge, and now two files sit at the same version. Nothing in the runner objects.

## Fix

`discover_migrations()` now raises on a collision, naming both files.

Failing at import is the lesser evil here: a duplicate version is a packaging mistake that takes seconds to fix once it is named, whereas a silently skipped migration costs hours and can leave data half-migrated in the meantime.

## Tests

`tests/test_migrations_runner.py` is new — there were no runner tests. Beyond the guard itself it includes a regression check that the real on-disk migration set has no duplicates, so a collision introduced later fails in CI rather than at a user's startup.

113 tests pass.

### DIFF
--- a/nerve/db/migrations/runner.py
+++ b/nerve/db/migrations/runner.py
@@ -12,10 +12,33 @@ import aiosqlite
 logger = logging.getLogger(__name__)
 
 
+def _check_no_duplicate_versions(items: list[tuple[int, str]]) -> None:
+    """Raise ``RuntimeError`` if two migration files share a version.
+
+    ``run_migrations`` skips anything at or below the recorded schema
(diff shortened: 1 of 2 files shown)


constkolesnyak:

PR #187, opened 2026-07-18
Render date placeholders in notifications deterministically

constkolesnyak:
Models in unattended cron sessions regularly miscompute weekdays for absolute dates — writing "24 June (Tue)" when 24 June is a Wednesday. Asking the model to be careful is unreliable, and embedding a lookup table in the system prompt does not scale past a couple of weeks. Nobody is reading along in a cron turn, so the wrong weekday ships.

This follows `print(f"{x:.2f}")` logic instead: the model declares semantic intent via a placeholder and the code formats it.

    <YYYY-MM-DD>           -> 24 June (Wed)
    <YYYY-MM-DD HH:MM>     -> 24 June (Wed), 19:00
    <2026-06-24>  (today)  -> today, 24 June (Wed)

The relative label arithmetic runs in Python from a known `now`, so the failure mode that motivated this cannot recur.

## Weekday placeholder

Source material often states a date as a bare weekday name with no calendar date — "arriving Tuesday". Resolving that is the same arithmetic the model gets wrong, with a second failure mode on top: rather than compute the weekday, the model reuses a date it saw earlier in the thread. So the model may copy the name verbatim and let the code resolve it:

    <dow:Tuesday>          -> nearest upcoming Tuesday, e.g. 21 July (Tue)

Malformed placeholders (impossible dates, out-of-range hours) are left untouched, so the model sees its own broken output rather than a silent miscarriage.

## Locale

Output language is `notifications.date_locale` — en (default), ru, de. Unknown or None falls back to English rather than raising.

Placeholder *parsing* is deliberately multilingual regardless of that setting: `<dow:Dienstag>` resolves even when rendering in English, because the source the model copied it from may be in any language. Only the rendered side is localized.

Rendering is applied at the single entry point shared by `send_notification`, `ask_question` and `propose_action`, all three with the configured locale.

72 tests, covering the default, both fallbacks, and locale-independent parsing.

### DIFF
--- a/nerve/config.py
+++ b/nerve/config.py
@@ -709,6 +709,11 @@ class NotificationsConfig:
         "high": "⚠️ ",
         "urgent": "🚨 ",
     })
+    # Output language for <YYYY-MM-DD> / <dow:> placeholders rendered into
+    # notification text: "en" (default), "ru", "de". Placeholder *parsing*
+    # stays multilingual regardless, since the source a weekday was copied
+    # from may be in any language. Unknown values fall back to English.
+    date_locale: str = "en"
 
     @classmethod
     def from_dict(cls, d: dict) -> NotificationsConfig:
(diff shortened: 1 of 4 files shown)


constkolesnyak:

PR #198, opened 2026-07-25
Add `nerve restart --resume` to continue a session across a restart

serxa:
## Problem

When a session restarts the daemon mid-turn (common while developing something that needs a restart), the in-flight turn is abandoned. Shutdown preserves the SDK conversation (`sdk_session_id`), but nothing re-issues the interrupted turn on startup — so the work silently stops until a new message arrives.

## Change

`nerve restart --resume <SESSION_ID>` (repeatable) enrolls sessions to be continued after the restart:

- The CLI appends the id(s) to an on-disk queue (`~/.nerve/resume-after-restart`) **before** triggering the restart, so the enrollment is durable and crash-safe (it survives even a hard kill).
- On startup the daemon drains the queue and drives one **internal** "continue where you left off" turn per still-resumable session, reusing the preserved `sdk_session_id` to restore full context — the same resume mechanism as `resume_session`. `internal=True` keeps the synthetic trigger out of the transcript while the assistant's continuation is persisted and broadcast normally.
- The queue file is deleted up front (so no id is ever re-run), and ids that no longer exist, are archived, are satellites (`source="external"`), or have no resumable SDK session are skipped.

There is no global switch or per-session policy: a session is resumed **iff** it was explicitly enrolled via `--resume`.

## Usage

```
nerve restart --resume <SESSION_ID>
```

## Testing

- New `tests/test_resume_after_restart.py` covers the drainer (parse / dedup / skip conditions / up-front drain / one-failure-isolation) and the CLI enqueue across modes.
- Full suite green (`.venv/bin/pytest tests/`); updated the MCP HTTP integration fixture to stub the new lifespan call.
- Backend-only — no frontend build needed.

### DIFF
--- a/nerve/agent/engine.py
+++ b/nerve/agent/engine.py
@@ -64,7 +64,7 @@ from nerve.agent.tools import (
 # keep working. The new runtime path uses ``self.registry`` + a
 # per-session ``ToolContext`` and ignores those globals.
 from nerve.agent.tools import init_tools
-from nerve.config import NerveConfig, load_mcp_servers
+from nerve.config import NerveConfig, RESUME_QUEUE_FILE, load_mcp_servers
 from nerve.db import Database
 from nerve.observability.langfuse import attributes as lf_attrs
 from nerve.skills.manager import SkillManager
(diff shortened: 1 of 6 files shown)


pufit:
cc @constkolesnyak

pufit (approved):

PR #221, opened 2026-07-29
chore: auth smoke test (ignore, will be closed immediately)

alex-clickhouse:
Verifying that the Nerve worker can push branches and open PRs after git/GitHub auth setup. Empty no-op commit. Closing right away.


alex-clickhouse:
Auth smoke test complete — closing.

PR #233, opened 2026-07-30
db: fail loudly when two migrations share a version number

constkolesnyak:
## What

Add a guard in the migration runner that raises an error if two migration modules use the same version number, instead of silently skipping or applying one.

Also adds a dedicated test file for the migration runner (`tests/test_migrations_runner.py`).

## Why

Silent duplicate version handling has caused subtle production issues — this guard surfaces the mistake immediately on startup.

### DIFF
--- a/nerve/db/migrations/runner.py
+++ b/nerve/db/migrations/runner.py
@@ -12,10 +12,33 @@ import aiosqlite
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


constkolesnyak:
Closing — I'm no longer maintaining patches against this repo; my fork now tracks upstream one-way. Feel free to pick up the change if it's useful.

PR #234, opened 2026-07-30
make systemd restart bounded so a self-restart cannot hang nerve down

constkolesnyak:
## What

When nerve restarts itself via `nerve restart`, the current implementation uses an unbounded helper script that can hang indefinitely if something goes wrong during startup. This change makes the restart bounded and systemd-aware: nerve signals systemd to respawn rather than launching an independent helper.

## Why

An unbounded self-restart can leave nerve in a stuck state with no recovery path. Making it systemd-aware lets the system's process supervisor handle failures and respawns correctly.

### DIFF
--- a/nerve/cli.py
+++ b/nerve/cli.py
@@ -412,15 +412,61 @@ def restart(ctx: click.Context, resume_ids: tuple[str, ...]) -> None:
         ctx.exit(rc)
         return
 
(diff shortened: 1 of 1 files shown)


constkolesnyak:
Closing — I'm no longer maintaining patches against this repo; my fork now tracks upstream one-way. Feel free to pick up the change if it's useful.

PR #237, opened 2026-07-30
fix anyio 4.13 CPU hot-loop: narrow zombie-scope patch

constkolesnyak:
## What

anyio 4.13 changed the internal `_deliver_cancellation` mechanism in a way that causes a CPU hot-loop (100% CPU) under certain task-group cancel patterns.

This patch (`nerve/_anyio_patch.py`) monkey-patches `_deliver_cancellation` at import time to:
1. Skip tasks that are already done (zombie-scope guard)
2. Narrow the scope so the patch only applies to the specific cancel pattern that triggers the loop

Includes comprehensive tests (`tests/test_anyio_patch.py`) that verify the patch doesn't regress the normal cancel paths.

## Why

100% CPU hot-loop causes nerve to become unresponsive. The patch is scoped narrowly to the problem scenario and includes tests for the fix and regression coverage for normal paths.

### DIFF
--- a/nerve/__init__.py
+++ b/nerve/__init__.py
@@ -1,3 +1,8 @@
 """Nerve — Personal AI Assistant."""
 
+# Apply anyio hot-loop patch as early as possible — monkey-patches
+# CancelScope._deliver_cancellation to prevent 100% CPU spins on
+# unrecoverable cancellations. See nerve/_anyio_patch.py for details.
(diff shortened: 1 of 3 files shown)


pufit (changes requested):
The fix was already merged upstream https://github.com/agronholm/anyio/issues/1111
Let's bump the version and close this PR

PR #254, opened 2026-08-03
Fix three memU write paths that damage store integrity

oranjeai:
## Symptom

Measured on a live 139,187-item store. `memory_update`, `memory_delete` and the web-UI
routes all reach these.

- **154 of 154** items ever updated without a `categories` argument now hold **zero** category
  links - about 47% of all 326 orphaned items.
- **149 of 149** updated items carry their *old* text's `content_hash` (400/400 never-updated
  items are fresh), so re-memorizing an item's pre-correction wording reinforces the corrected
  row rather than creating a new one.
- **6,455 dangling `memu_category_items` rows**, all 5,611 distinct `item_id`s present in the
  `item_deleted` audit log; they inflate `memory_expand_category`'s `total` by 3.5-4.6% per
  category (relations counted, items listed through a JOIN).

## Root cause

1. memU has **one sentinel for two meanings**: an omitted `categories` becomes `None`, and
   `_map_category_names_to_ids(None, ctx)` returns `[]`, so `_patch_update_memory_item`'s
   `cats_to_remove` is the item's *entire* current set, each unlink recording
   `(old_content, None)` - read downstream as "discarded", so the LLM drops the item from the
   category summary too.
2. `content_hash` is a **derived field with no invalidation**: `update_item` rewrites
   `summary`/`memory_type` and never touches it, while `create_item_reinforce` dedups on it.
3. **No layer owns the dependent rows**: `delete_item` deletes only the item row; there is no
   FK or cascade.

## The change

Four patches in `_patch_sqlite_bugs()`, the established seam for this class of memu-py defect
(precedent: #119).

- **Fix 8** - when `categories` is `None`, neutralise membership for **that one call**: the
  delegated handler gets a per-call proxy whose `link`/`unlink` are no-ops, so its diff removes
  nothing, and `(old, new)` is rebuilt from the surviving links. Nothing process-wide is
  mutated. Explicit list still replaces; explicit `[]` still clears.
- **Fix 9** - recompute `content_hash` from the **DB row** when `summary`/`memory_type`
  changes. Only an existing hash is refreshed.
- **Fix 10** - delete relations and the item in **one** transaction, so a failure can never
  leave a surviving item stripped of its links. Installed before Fix 3 so its vector-index hook
  still wraps it.
- **Fix 7 correction** - the dedup writeback wrote a *cache*-derived `extra` over the row's, and
  read paths omit `extra=`, so a reinforce through a cold cache deleted `content_hash` and any
  key another writer had added. It now seeds `extra` from the row.

## Validation

52 new tests. Each defect has a control arm driving memU's own unpatched function against an
identically-built fixture, so a test passing without the fix is a failed test, not a passing
fix. A 38-arm mutation matrix kills 35, each naming a failing test, no-op controls green at both
ends; the 2 survivors are unreachable by construction, measured. Full suite **2985 passed**,
pre-existing failure set unchanged by name.

<details>
<summary>Scope, unfixed siblings, behaviour changes</summary>

**Write paths only - no data is repaired.** The store still holds 6,455 dangling relations, 326
orphans, 149 stale hashes and 6,473 hashless rows; the 154 lost memberships are **not
recoverable** (the audit log records `categories_changed: false`, not the ids).

The Fix 7 wipe explains that population: all 6,473 have `rc > 1` and none is a `tool` item, the
only create path that legitimately writes no hash.

Both fixes read the row, not the cache: a cached item has `extra == {}`, so a cache-based refresh
refreshes nothing in a long-lived process.

**Pinned, not closed.** Fix 9's CAS binds `summary`/`memory_type`, so any hash it writes is the
row's own, but the refresh is best-effort: if skipped (a lock) the row keeps its old hash, as
base always does - logged and asserted. A writer landing between memU's write and this read can
likewise leave the *cache entry* on that writer's hash, unread (dedup reads the DB). Fix 7 keeps
memU's read-modify-write shape, so a key written inside it is still lost (base loses every key).
All pinned by tests; a CAS is tracked apart.

**Named but deliberately unfixed.** Test-pinned so the exemption cannot rot: read paths still
omit `extra=` (no longer destructive, so hydration is its own change); `clear_items` has the same
dangling shape but no nerve callers; `PatchMixin` is dead code. Unreachable and untested:
the in-memory/postgres siblings (provider hardcoded `sqlite`).

**Behaviour changes.** A reinforce preserves more of `extra`, including `ref_id`, which
`list_items_by_ref_ids` filters on. Hash-dedup now matches rows whose hash used to be wiped, so
some memorizations reinforce rather than duplicate (the intent). Recomputing the 149 stale
hashes collides with none of the 132,714 hashed rows; updating an item's text to exactly match
another's now yields two rows sharing a hash, handled as the dedup lookup handles two
identical `memorize` calls.

</details>

### DIFF
--- a/nerve/memory/memu_bridge.py
+++ b/nerve/memory/memu_bridge.py
@@ -1032,16 +1032,22 @@ class MemUBridge:
                             self._memory_item_model.id == item_id
                         )
                     ).first()
+                    # Relations are deleted on BOTH paths: when the item row is
+                    # already gone (another process removed it) its
+                    # memu_category_items rows are exactly the dangling rows
+                    # this fix exists to prevent, so skipping them there would
+                    # leave the changelog contract half-kept.  Deleting the
+                    # item is still conditional -- an unknown id stays a
+                    # no-op, since the relation DELETE matches nothing.
+                    session.exec(_del(rel_model).where(rel_model.item_id == item_id))
                     if row is not None:
-                        session.exec(_del(rel_model).where(rel_model.item_id == item_id))
                         session.delete(row)
-                        session.commit()
+                    session.commit()
 
                 # Evict on BOTH paths, like memU's own delete_item.  When the row
                 # is already gone (another process removed it) returning early
                 # would leave the id in self.items, and Fix 5 serves that cache
                 # unfiltered, so list_items() would keep returning a deleted item.
-                # No DB write is attempted for an absent row.
                 self.items.pop(item_id, None)
                 relations = self._state.relations
                 relations[:] = [r for r in relations if r.item_id != item_id]
@@ -1522,18 +1528,59 @@ class MemUBridge:
             # returns [] for a falsy list), so cats_to_remove becomes the item's
             # ENTIRE current set and each link is unlinked.
             #
-            # An omitted `categories` therefore performs NO membership mutation at
-            # all: the delegated handler runs with link/unlink neutralised, so no
-            # link can be removed by a diff it never should have computed.  The
-            # (old, new) pairs the LLM category-summary step consumes are rebuilt
-            # here from the links that actually survive.
+            # An omitted `categories` therefore performs no membership mutation:
+            # the delegated call sees a relation repo whose link/unlink are
+            # no-ops, so no link can be removed by a diff it never should have
+            # computed.  The shared repo is never modified -- the no-ops live on
+            # a proxy built per call and reachable only through the `store` this
+            # call passes down, so a concurrent coroutine on the same memU loop
+            # keeps the real mutators.  The (old, new) pairs the LLM
+            # category-summary step consumes are rebuilt here from the links
+            # that actually survive.
             from memu.app.crud import CRUDMixin as _CRUDMixin
 
             if not getattr(_CRUDMixin._patch_update_memory_item, "_nerve_keeps_categories", False):
                 _pre_keep_update_handler = _CRUDMixin._patch_update_memory_item
                 _CRUDMixin._nerve_memu_update_handler = _pre_keep_update_handler
-                _KEEP_MISSING = object()
-                _KEEP_MUTATORS = ("unlink_item_category", "link_item_category")
+
+                class _KeepRelRepoProxy:
+                    """Per-call relation repo whose membership mutators are no-ops.
+
+                    Every other attribute is forwarded to the real repo, so the
+                    delegated handler's reads (get_item_categories) see live
+                    data.  Instantiated inside the call, so nothing shared is
+                    written and there is no restore to get wrong.
+                    """
+
+                    __slots__ = ("_nerve_real",)
+
+                    def __init__(self, real):
+                        object.__setattr__(self, "_nerve_real", real)
+
+                    def unlink_item_category(self, *_args, **_kwargs):
+                        return None
+
+                    def link_item_category(self, *_args, **_kwargs):
+                        return None
+
+                    def __getattr__(self, name):
+                        return getattr(object.__getattribute__(self, "_nerve_real"), name)
+
+                class _KeepStoreProxy:
+                    """Per-call store serving the relation proxy, nothing else changed."""
+
+                    __slots__ = ("_nerve_real", "_nerve_rel")
+
+                    def __init__(self, real, rel_proxy):
+                        object.__setattr__(self, "_nerve_real", real)
+                        object.__setattr__(self, "_nerve_rel", rel_proxy)
+
+                    @property
+                    def category_item_repo(self):
+                        return object.__getattribute__(self, "_nerve_rel")
+
+                    def __getattr__(self, name):
+                        return getattr(object.__getattribute__(self, "_nerve_real"), name)
 
                 async def _category_preserving_update(self, state, step_context):
                     payload = state.get("memory_payload") or {}
@@ -1547,31 +1594,34 @@ class MemUBridge:
                     item_before = store.memory_item_repo.get_item(memory_id)
                     old_content = getattr(item_before, "summary", None)
 
-                    def _keep_noop(*_args, **_kwargs):
-                        return None
+                    # Neutralise membership for THIS call only.  The delegated
+                    # handler reads store.memory_item_repo and
+                    # store.category_item_repo; the proxies serve both.
+                    store_proxy = _KeepStoreProxy(store, _KeepRelRepoProxy(rel_repo))
+                    out = await _pre_keep_update_handler(
+                        self, {**state, "store": store_proxy}, step_context,
+                    )
 
-                    saved = {
-                        name: rel_repo.__dict__.get(name, _KEEP_MISSING)
-                        for name in _KEEP_MUTATORS
-                    }
-                    for name in _KEEP_MUTATORS:
-                        setattr(rel_repo, name, _keep_noop)
-                    try:
-                        out = await _pre_keep_update_handler(self, state, step_context)
-                    finally:
-                        for name, prev in saved.items():
-                            if prev is _KEEP_MISSING:
-                                rel_repo.__dict__.pop(name, None)
-                            else:
-                                rel_repo.__dict__[name] = prev
+                    # run_steps threads the returned mapping into persist_index
+                    # and build_response, so the proxy must not escape its
+                    # window: hand the REAL store back.
+                    out = dict(out)
+                    out["store"] = store
 
                     # memU diffed against a no-op, so its pairs are wrong: a live
                     # link marked (old, None) renders as "content is discarded".
                     if payload.get("content"):
                         new_content = getattr(out.get("memory_item"), "summary", None)
+                        # Keep only ids the response step can resolve:
+                        # _patch_build_response subscripts
+                        # memory_category_repo.categories UNGUARDED, so a
+                        # dangling relation row would raise KeyError after the
+                        # content write has already committed.
+                        known = store.memory_category_repo.categories
                         out["category_updates"] = {
                             rel.category_id: (old_content, new_content)
                             for rel in rel_repo.get_item_categories(memory_id)
+                            if rel.category_id in known
                         }
                     else:
                         out["category_updates"] = {}
--- a/nerve/memory/memu_bridge.py
+++ b/nerve/memory/memu_bridge.py
@@ -1101,59 +1101,97 @@ class MemUBridge:
                     tool_record=tool_record,
                 )
 
-                if memory_type is None and summary is None:
+                # Everything below is BEST-EFFORT.  It runs after memU's content
+                # write has already committed, so a failure here (a lock, say)
+                # must not turn a landed update into a raising call: that would
+                # leave the content written but the categories undiffed
+                # (_patch_update_memory_item diffs AFTER update_item), the vector
+                # index on the old embedding, and bridge.update_item reporting
+                # False for an update that partly applied.  Degrading to a stale
+                # hash is exactly base's unconditional behaviour, so it is the
+                # strictly safer failure.  The delegation itself stays OUTSIDE
+                # this guard: a genuine update failure must still propagate.
+                try:
+                    if memory_type is None and summary is None:
+                        return result
+
+                    with self._sessions.session() as session:
+                        row = session.exec(
+                            _sel_upd(self._memory_item_model).where(
+                                self._memory_item_model.id == item_id
+                            )
+                        ).first()
+                        if row is None:
+                            return result
+                        current = dict(row.extra or {})
+                        # Only refresh a hash that already exists: an item
+                        # created without one must not be newly enrolled into
+                        # hash-dedup.
+                        if not current.get("content_hash"):
+                            return result
+                        want = _content_hash(row.summary, str(row.memory_type))
+                        # The only interpolation is the model's own __tablename__
+                        # (Fix 6 renames memu's tables, so it cannot be a
+                        # literal); every value is a bound parameter.  RETURNING
+                        # hands back the row's own merged extra, so the cache is
+                        # never rebuilt from the pre-write snapshot.
+                        returned = session.execute(
+                            _sql_text(
+                                f"UPDATE {self._memory_item_model.__tablename__} "
+                                "SET extra = json_set("
+                                "coalesce(extra, '{}'), '$.content_hash', :want) "
+                                "WHERE id = :item_id "
+                                "AND json_extract(extra, '$.content_hash') IS NOT NULL "
+                                "AND summary = :summary AND memory_type = :memory_type "
+                                "RETURNING extra"
+                            ),
+                            {
+                                "want": want,
+                                "item_id": item_id,
+                                "summary": row.summary,
+                                "memory_type": str(row.memory_type),
+                            },
+                        ).fetchall()
+                        session.commit()
+                        # rowcount is not meaningful with RETURNING; the row
+                        # count is.  Empty means the CAS declined: another writer
+                        # moved summary/type, or removed the hash, between the
+                        # read above and this statement.  Nothing was written, so
+                        # claiming a refreshed hash in the cache would make it
+                        # disagree with the row.
+                        if not returned:
+                            return result
+                        # json_set merges into the row's CURRENT extra while the
+                        # CAS binds only summary/memory_type, so a writer that
+                        # changed any OTHER key in the read -> write window
+                        # committed and is present here.  Reconstructing
+                        # {**current, ...} would silently revert that key in the
+                        # cache and the returned item - including
+                        # reinforcement_count, which Fix 3's salience ranking
+                        # reads.  The JSON column may hand back either a string
+                        # or an already-decoded dict.
+                        written = returned[0][0]
+                        if isinstance(written, str):
+                            written = json.loads(written)
+                        written = dict(written or {})
+
+                    # Keep the cache consistent with what was written, like the
+                    # Fix 7 writeback does.
+                    cached = self.items.get(item_id)
+                    if cached is not None:
+                        cached.extra = written
+                    if getattr(result, "id", None) == item_id:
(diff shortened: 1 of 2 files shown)


oranjeai:
<details>
<summary>Internal second-model review - 9 Gate B rounds, 55 findings adjudicated</summary>

Every change here was reviewed by a second model with no access to my reasoning, then
adjudicated against the code. Rounds are uncapped; this took nine, and the last five each
closed a defect the previous round had left behind.

### Blockers found and fixed

| # | Finding | What was wrong | Fix |
|---|---|---|---|
| 1 | `fix8-stale-snapshot` | Fix 8 rebuilt the category payload from a snapshot read before the write | mechanism replaced |
| 2 | `own-r4-stub-window-is-process-wide` | the replacement stubbed `link`/`unlink` on a **process-wide** repo instance across an `await`, so any concurrent pipeline's membership write was silently swallowed, and the restore was not reentrant | replaced again, with a per-call proxy reachable only through the `store` that call passes down |
| 3 | `own-r4-fabricated-category-updates-keyerror` | the rebuilt `category_updates` fed raw relation rows to a consumer that subscripts the category map **unguarded**, after the content write commits | filter to resolvable ids |
| 4 | `hash-writeback-atomicity` | the hash refresh could stamp a hash for text the row no longer held | conditional `UPDATE` bound to `summary`/`memory_type`, new extra taken from `RETURNING` |
| 5 | `category-noop-race` | the no-op window was observable by a concurrent caller | per-call proxy (same fix as #2) |
| 6 | `stale-row-ghost` | on a matched row another process deleted, the reinforce path incremented the cached copy and returned it with no DB write | **reverted from this PR** - see below |

### Disagreed, with evidence

- **`fix8-cold-category-cache`** (this round's only finding, a blocker). Claim: with a cold
  category cache the filter in #3 drops a live id, so the persisted category summary stays
  stale. Refuted on a **necessary step**, with a control: memU's summary step reads the *same*
  cache the filter reads, via `.get(cid)` + `continue` (`crud.py:651-653`, again at `:666`).
  Modelled both arms - id uncached with the filter **removed** (so the id *is* in
  `category_updates`): `update_category` calls **0**; same id cached (control): **1**. In the
  cold state memU itself skips the summary whether or not the filter is present. Two further
  refutations: the response payload the filter shapes has **no consumer** (the bridge awaits
  the call and returns a bool, never reading `category_updates`); and the filter is
  load-bearing - removing it re-admits a `KeyError` from an *unguarded* subscript that runs
  **after** the content write commits, verified to raise for an id absent from the cache, on a
  store measured today to hold 6,455 dangling relations over 5,611 distinct item ids. The
  mechanism half is accepted as true (`known` is the cache, not the database); it is simply not
  a defect this PR introduces or can fix at that line.
- **`hash-refresh-best-effort` / `hash-refresh-two-phase` / `fix9-best-effort-hash`** (raised
  four times). The prescribed remedy - make the content write and the hash refresh atomic - was
  **built** in an earlier round and measured to **regress two correctness tests** by storing a
  wrong hash, because memU's own write is what commits the content. Refused on that
  measurement, not on scope. The refresh stays best-effort and post-commit: if skipped, the row
  keeps its old hash, exactly today's behaviour. Logged, and asserted by a test that injects
  the lock.
- **`date-sweep-clobbers-extra`** (raised twice). Mechanism **confirmed** and real, but its
  carrier is a different code path, byte-identical at the base, and outside this changelog.
  Escalated separately rather than absorbed here.
- **`delete-link-race`** (raised three times). Mechanism real, but the inserting statement is
  memU's and this PR never modifies it, and at the base every delete dangles with no race at
  all. Already investigated to completion elsewhere, where the measurement was that **0 of
  6,455** live dangling rows come from that path.
- **`reinforce-extra-lost-update`**. Mechanism confirmed; the base performs the same
  read-modify-write, so this PR only shrinks the loss. The residual window is pinned by a test
  that first proves the interleaving happened, and disclosed in the body.

### Reversals and corrections I made to my own work

- **`stale-row-ghost` is REVERTED in the final revision.** I agreed it in an earlier round and
  wrote a fix plan for it; that was **my error**. The identical remedy was already open as
  #248, created over three hours before the round that asked for it here, and carrying it in
  two PRs of mine would force a hand resolution on whichever merged second. The reverted
  functions are byte-identical to their pre-revert parent (AST extraction: 57 and 36 lines,
  with the same comparison against the reverted revision reporting 57 vs 76, so the comparator
  discriminates). #248 carries the fix and five tests asserting both the cache and the index
  eviction, verified rather than assumed. No fix is lost.
- **My own validation comment claimed that reverted mechanism as shipped.** Caught in this
  round's cold review and corrected before posting: the a-i comment's invariants row described
  the eviction as delivered, while the commit message on the same PR correctly said it was
  reverted. Measured on the shipped tree - zero occurrences of the eviction in the whole diff.
- **An exemption was described as comment-pinned when the same revert deleted that comment.**
  Corrected on both surfaces: three of the named exemptions are pinned by tests, and the
  hardcoded-`sqlite` provider one is pinned by **no** test - stated plainly rather than
  implied.
- Two inherited mutation-matrix arms were found **mis-sited** by substring matching, so one
  recorded kill had been **vacuous**. The mutator now asserts every anchor matches at a line
  boundary, proven in both directions.
- Earlier bodies and commit messages carried figures that later rounds made stale (44, then 51,
  then 53 tests). Every figure in the final body was re-derived from the shipped tree, three
  independent ways where a count was involved.

### Validation posture

52 targeted tests over five write-path classes, none of which exists at the base. A 38-arm
mutation matrix kills 35, each naming a failing test, with no-op controls green at both ends
and zero vacuous arms; the two survivors are unreachable by construction and were re-measured
on this tree rather than carried, and three arms whose subject the revert removed are retired
**with cause**, printed by name. Full suite 2985 passed, failure set identical by name to this
revision's own pre-edit baseline.

**Known residual, disclosed rather than resolved:** this branch conflicts with #248 in both
files, and did so from its first commit - before the reverted change existed (measured at every
commit: clean at the merge base, conflicting from the first commit on). The overlap is the
surrounding `extra`-seeding edit, not the reverted eviction. Not resolved here, because #248 is
unmerged and stacking on an unmerged prerequisite is its own hazard. #249 and #251 also
conflict, in the test file only, where all three append test classes at the end.

Session id: cron:clickhouse-review-slot-11:20260803-181700

</details>

oranjeai:
### Pre-PR validation gate

<details>
<summary>a-i self-check, with the measurement behind each answer</summary>

| # | Question | Answer |
|---|---|---|
| a | **Deterministic repro?** | YES. A self-verdicting 14-arm script run against an unmodified `main` export (`git archive`, no `.git`) reports **9 arms DEFECT**; against this branch **14/14 OK**. 100% reproducible, not a percentage. Note memU is not constructible without the existing Fixes 1/2/6 (`TypeError: Cannot create a consistent MRO`, and memu's own `sqlite_*` table names are reserved by SQLite), so the control is `_patch_sqlite_bugs()` as it stands on `main`, i.e. production behaviour today. |
| b | **Root cause explained?** | YES. One sentinel for two meanings (`crud.py:558-566` + `:625-626`); a derived field with no invalidation (`memory_item_repo.py:311`/`:315-317` vs `:388-460`, where `grep -c content_hash` = 0); no layer owning the dependent rows (`crud.py:593` -> `:461-475`, and the `memu_category_items` DDL has no FK); and cache-as-canonical in the Fix 7 writeback, since all four read paths (`:73`, `:104`, `:149`, `:183`) build `MemoryItem` without `extra=` while all four write paths pass it. |
| c | **Fix matches root cause?** | YES. Each patch sits in the function that owns the broken decision: the update handler that owns "was `categories` supplied?", the repo method that owns both the source columns and the derived field, the delete that owns its dependent rows, and canonical state read inside the write transaction. Fix 8's neutralisation is scoped to ONE CALL: a proxy built inside the call and reachable only through the `store` that call passes down, never a mutation of the process-wide repo instance, whose no-ops would otherwise be in force for every concurrent coroutine for the duration of the delegated `await`. No widened bounds, no guard over an unfixed cause. Two alternative shapes were rejected during planning **with measurements** - compensating after delegation leaves a `(old, None)` "discarded" signal, and a `get_item()`-based hash refresh is a production no-op - and mutants M2/M5 keep both rejections enforced. |
| d | **Test intent preserved / new tests?** | YES. No existing test weakened. One pre-existing test (`TestIndexedUpdateItemForwarding`) needed its global-state restore list widened and two attributes on its stub; **its assertions are byte-unchanged**. **52 new tests** - none of the five write-path classes exists at `origin/main`, so every test in them is new; counted three ways (added test defs per class, pytest's own selection count, and the classes' absence at the base) all agreeing. This figure SUPERSEDES the earlier "44", "51" and "53"; re-measured three ways this round, all agreeing at 52 (14 + 19 + 6 + 7 + 6 per class; collected 128 vs 76 at the base; 52 added `def test_` lines). This round REMOVES one case - the ghost-match eviction test, whose contract belongs to the already-open PR #248 that this round reverts a duplicate of - and keeps the case pinning the residual hash window described in row g. Nine tests the fix plan required unchanged were verified **byte-identical** by extracting each function body and comparing against the previous revision. |
| e | **Both directions?** | YES, three independent ways: the 9-arm delta above; per-defect control arms **inside** the suite that drive memU's own unpatched function against an identically-built fixture (`test_unpatched_handler_loses_every_link`, `test_unpatched_update_leaves_the_hash_stale`, `test_unpatched_delete_orphans_the_relations`); and a **38-arm** mutation matrix (nothing carried blindly - the inherited arms are IMPORTED from the previous round's mutator and every anchor is asserted to resolve exactly once against this tree before the run), in which **35 arms are killed** each naming at least one failing test while the no-op control stays green at BOTH ends (0 vacuous, tree verified restored after every arm, and every export asserted to CARRY this round's edits). The two survivors are unobservable BY CONSTRUCTION and measured rather than assumed, RE-MEASURED on this tree rather than carried: one drops a proxy override on a code path proven unreachable (instrumented live, `_map_category_names_to_ids` is called once with `None` and returns `[]`, so `cats_to_add` is empty and memU's link loop never runs), the other reads a mapping the proxy forwards by object identity (observed from inside the delegated call: `memory_category_repo is` the real repo True, `.categories is` the same object True, and `category_item_repo is` real False - the one override). The three arms an earlier revision added for the ghost-eviction branch are RETIRED WITH CAUSE this round (printed by name at run time, never dropped silently): that branch is reverted here as a duplicate of PR #248, so their subject no longer exists in this PR, and their coverage now lives in #248 - verified rather than assumed, since #248 ships 5 tests asserting both the cache and the index eviction. The two arms that revision had re-anchored are DE-re-anchored: the revert restores their block byte-identically, so the r5 anchors are imported verbatim and all 38 are asserted to resolve exactly once. The harness also asserts NEGATIVE markers - the reverted text must be ABSENT from every export - so no arm can silently re-test pre-revert code. Two INHERITED arms were also found MIS-SITED - their whitespace-leading anchors matched more deeply indented lines, so a hit count of 1 did not prove where they landed, and one of them therefore recorded a **vacuous kill** in the previous revision; both are re-anchored and now genuinely kill, and the mutator asserts every anchor match begins at a line boundary (proven to catch both original anchors while passing all 38 current arms). The matrix earned its cost four times over: substituting `extra` instead of merging silently dropped a caller's `ref_id`; taking the hash's summary or type from the ARGUMENT rather than the row passes every non-racing case; a present-but-EMPTY hash is the one input where the Python guard and the SQL predicate disagree; and my own first version of the hash writeback stamped the cache even when the conditional write declined. Each is now pinned by a test. This round the matrix found the sharpest one yet: the ERROR CONTRACT of the hash phase was entirely unobserved - a mutant wrapping that whole phase in `try/except Exception: return result` was indistinguishable from the propagating shape, passing 39 of 39, so either contract satisfied the suite. Two cases now pin it in both directions. The hash-refresh restructure is also proven to discriminate against the shape it replaces: run against the previous two-session implementation, the new race test fails on its CORE assertion. |
| f | **General across code paths?** | YES. All three nerve call sites (agent tool, the bridge's own `knowledge_filter` delete, the web-UI routes) go through the patched bridge methods. Every sibling carrier is either fixed or named with the measurement that makes it unreachable: provider is hardcoded `sqlite` so the in-memory/postgres repos and the in-memory reinforce arm cannot be the backend; `clear_memory` has 0 nerve callers; `PatchMixin` has no subclass and is absent from `MemoryService.__mro__`. Tests pin the last two facts (`test_clear_items_remains_unreachable_from_nerve`, `test_patch_mixin_duplicates_are_dead_code`); the hardcoded-`sqlite` provider fact is pinned by NO test (`grep -c provider tests/test_memu_bridge.py` = 0) and this revision's revert also removed the source comment that named it, so that one exemption now rests on review alone. It is a `MemoryService` config value at `memu_bridge.py:1864`, changeable without touching either patched file. |
| g | **Generalizes across inputs?** | YES. All four configured memory types refresh correctly. Boundary summaries - empty, length 1, unicode, whitespace-only, 20,000 chars - all refresh correctly. `categories` as `None` / an explicit list / an explicit `[]` / an item with zero links all behave as intended, and the update SHAPE matrix covers content-only, type-only and content+type, the last now asserted directly - preservation does not depend on `content` being supplied (a mutant gating it on content survived 31 tests before a type-only case was added), and a conditional mutant that keeps the OLD type whenever a summary is supplied survived 36 tests before the combined case was added. Hash enrolment is measured across all three `content_hash` states: absent, JSON null, and present-but-empty. A missing id raises memU's own `KeyError`, unchanged. A non-round-tripping category map fails closed with every link still present. That former residual window is now CLOSED and pinned closed: membership is not mutated at all, so a category link inserted by raw SQL between the two reads survives and is reported to the summary step (`test_a_link_added_between_the_two_reads_is_preserved`). A different window IS disclosed rather than claimed closed: a third-party `extra` key written between the Fix 7 writeback's SELECT and its flush is lost, which base also loses (base loses every key, with no race at all), pinned by `test_a_key_written_inside_the_writeback_window_is_still_lost` and left for a separate CAS decision. A relation row naming a category id the response step cannot resolve is filtered out of `category_updates`, because memU subscripts that mapping unguarded AFTER the content write commits. |
| h | **Backward compatible?** | YES. `git diff --name-only main` is exactly two files; no schema, migration, config key, dependency or tool-schema change. Nothing depends on the destructive behaviour: none of the four call sites passes `categories` expecting a clear, and the tool handler's own "Nothing to update" guard shows the intent was always "update what was given". The three real behaviour changes are stated in the PR body rather than slipped in. |
| i | **Invariants preserved?** | YES. The invariant is "a write must not destroy state the caller did not ask to change, and must leave every derived artifact consistent". Verified across a create+update+delete+update sequence: 2 items, 6 relations, **0 dangling**, and both in-process caches exactly matching the DB. Error paths covered in BOTH directions: forcing the item delete to raise leaves the relation rows AND the item row intact, and forcing only the relations delete to raise leaves the item row intact too (one session, since `SQLiteSessionManager.session()` returns a fresh `Session` per call). The second direction is the one that discriminates: `session.delete` raises first in both a one-transaction and an item-first-split form, so an item-first split survived 31 tests until a relations-only failure case was added. This PR does NOT change the semantic-dedup ghost path: an earlier revision evicted the stale cache and vector-index entry when the matched row was gone, and this revision REVERTS that, because the same remedy is already open as PR #248 (created three hours before the round that asked for it here). `_semantic_sqlite_reinforce` and its in-memory sibling are byte-identical to this branch's pre-revert parent, verified by AST extraction (57 and 36 lines; the same comparison against the reverted revision reports 57 vs 76, so it discriminates). Base's outcome on that path is unchanged by this PR, and #248 carries the fix plus five tests asserting both the cache and the index eviction. A missing id is a no-op for delete - the relation DELETE simply matches nothing - and a row another process already deleted still gets its cache entry and cached relations evicted AND its relation rows removed from the DB, asserted by raw count: on that path those rows ARE the dangling rows this fix exists to prevent, so the relation delete is no longer conditional on the item row (a mutant restoring that condition is killed by exactly the new raw-DB assertion). Fix 8 no longer raises at all, so there is no refused-update path left to reason about. The hash refresh holds the same invariant on EVERY exit path, including the one this round added: when its conditional write declines because another writer moved the row, nothing is written and the cache is left agreeing with the row rather than claiming a refresh that never landed (measured: db hash untouched, cache hash equal to it; a mutant dropping that gate leaves them disagreeing). No concurrency or lock-ordering contract is touched, and the added read is one indexed primary-key select on a rare human-initiated operation. The hash phase is also best-effort by contract, and that contract is now pinned rather than implied: it runs after memU's content write has committed, so a failure there leaves the row with a stale hash (base's unconditional behaviour) and a WARNING, while memU's own update call stays outside the guard so a genuine update failure still propagates. Both directions are asserted, and the shipped propagating shape is killed by the first case. Finally, the refreshed `extra` is taken from the write itself via `RETURNING` rather than rebuilt from the pre-write snapshot: `json_set` merges into the row's current `extra` while the write's predicate binds only summary/memory_type, so a concurrent writer's key (including `reinforcement_count`, which the salience ranking reads) survives in the row and must not be reverted in the cache. Measured both ways. |

Also checked: the vector-index hook stays outermost and still fires; repeating
`_patch_sqlite_bugs()` does not double-wrap (verified over three consecutive calls); the
`reinforcement_count > 1` decision that gates category linking after a reinforce is
**unchanged** - had it flipped, reinforced items would have stopped getting their links, i.e. a
new source of exactly the orphans this fixes.

</details>

oranjeai:
cc @pufit @serxa — could you review this? It patches three memU write paths in `_patch_sqlite_bugs()`: a content-only `memory_update` currently unlinks every category (memU maps a missing `categories` to `[]`, so its diff removes the item's whole current set), `update_item` never recomputes `extra.content_hash` that `create_item_reinforce` dedups on, and `delete_item` leaves its `memu_category_items` rows behind. Measured on the live store: 154/154 items updated without `categories` hold zero links, 149/149 updated rows carry their old text's hash, 6,455 dangling relations.

PR #259, opened 2026-08-03
memory: seed the categories memU actually advertises

oranjeai:
## Symptom

With no `memory.categories` in config, the memorize LLM is told 10 categories exist while the
store holds none, so every name it emits is silently dropped and category recall stays
permanently empty. Measured on an empty DB and config: advertised 10, rows 0, map 0, resolvable
**0 of 10**. That is the default: `config.example.yaml` ships no `categories` key.

## Root cause

nerve seeded from `config.memory.categories`, memU advertised from `service.category_configs`;
they coincide only when nerve configures its own. Otherwise memU's 10 defaults reach the prompt,
`_ensure_categories` early-returned, and `categories_ready=True` suppressed memU's initializer.
The `list_categories()` warming the rebuild's cache sat behind that return too, so a
runtime-created category was unresolvable after restart.

## The fix

Seed from `service.category_configs` and load persisted rows unconditionally. Seeding writes
through the category repository, not `_create_category_impl`: that also appends a
`CategoryConfig` for an already-advertised name, doubling the advertised set -- **live today on
the configured path too**.

Embeddings are computed in one batched call before any write, and a failure propagates: a
null-embedding row is never repaired, so a `None` fallback would trade one silent failure for a
permanent one.

`_available` moves to the end of `_initialize_impl` from ~290 lines early, before seeding:
`engine.py` discards `initialize()`'s return, so it is the agent's only failure signal.

Behaviour changes: a default install creates 10 rows on first boot (idempotent by name); a row
whose stored name is not memU's form is renamed in place, keeping id and items, unless another
row already owns its lookup key; and an embedding outage or a category read error now fails
memory init loudly. Out of scope: null-embedding rows (#70); runtime-created categories stay
unadvertised.

## Validation

44 new tests, 35 failing against unmodified source (41 at file scope, six pre-existing).
Suite 2933 -> 2977 passed, identical 7-name pre-existing failure set.

### DIFF
--- a/nerve/memory/memu_bridge.py
+++ b/nerve/memory/memu_bridge.py
@@ -1792,7 +1792,7 @@ class MemUBridge:
             try:
                 self._service.database.memory_item_repo.list_items()
                 # Categories are already cached: _ensure_categories loads them
-                # unconditionally, before the name-to-ID rebuild above.
+                # whenever a service exists, which _initialize_impl guarantees.
 
                 # Convert cached embeddings from list[float] to numpy float32.
                 # Pydantic coerces numpy → list during model construction, so we
--- a/nerve/memory/memu_bridge.py
+++ b/nerve/memory/memu_bridge.py
@@ -123,6 +123,34 @@ def _category_breadcrumb(name: str, description: str, summary: str) -> str:
     return text
 
 
+def _memu_cat_name(name: str) -> str:
+    """The display name memU advertises and stores: ``name.strip() or "Untitled"``.
+
+    ``memu/app/memorize.py:930-938`` (prompt) and ``:663-665`` (persist).  Not
+    lowercased: lowercasing happens only at lookup.
+    """
+    return name.strip() or "Untitled"
+
+
+def _memu_cat_key(name: str) -> str:
+    """The key memU's own resolver computes: ``name.strip().lower()``.
+
+    ``memu/app/memorize.py:682``.  The name-to-id rebuild in ``_initialize_impl``
+    does NOT strip, so occupancy is judged in the rebuild's domain, not here.
+    """
+    return _memu_cat_name(name).lower()
+
+
+def _memu_cat_embed_text(name: str, description: str) -> str:
+    """memU's ``_category_embedding_text`` (``memu/app/memorize.py:670-673``).
+
+    One copy, so seeds and repairs share the space ``cosine_topk`` ranks in.
+    """
+    desc = (description or "").strip()
+    name = _memu_cat_name(name)
+    return f"{name}: {desc}" if desc else name
+
+
 class _VectorIndex:
     """Incremental brute-force vector index over memU item embeddings.
 
@@ -1881,53 +1909,111 @@ class MemUBridge:
         # Unconditional, and read failures propagate: this call also warms the
         # repo cache that _initialize_impl's name-to-ID rebuild reads, and an
         # ``existing`` left empty by a swallowed error would re-seed every name.
-        existing = {getattr(cat, "name", "") for cat in repo.list_categories().values()}
+        rows = list(repo.list_categories().values())
 
+        # Group by the LOOKUP key, not the display name: the name-to-id rebuild in
+        # _initialize_impl and memU's own resolver agree on it up to stripping, so
+        # a rename that is free among display names can still collapse two live
+        # rows onto one map entry.
+        key_owners: dict[str, list[Any]] = {}
+        for cat in rows:
+            key_owners.setdefault(_memu_cat_key(getattr(cat, "name", "")), []).append(cat)
+
+        # Repair rows persisted under a non-normalized name so their rebuild key
+        # becomes the one memU computes.  Two live sources: an upgrade from the
+        # pre-fix code, which stored the raw configured name, and
+        # ``_create_category_impl``, the runtime path this change leaves alone.
+        # The row id survives the rename, so its category_items stay linked.
+        # A key owned by more than one row is left entirely alone -- renaming
+        # either would make them indistinguishable to every lookup, and merging
+        # would have to discard one row's items.  Read from the whole set, not
+        # the rows walked so far: list_categories() applies no ORDER BY.
+        repairs: list[tuple[Any, str]] = []
+        for cat in rows:
+            raw = getattr(cat, "name", "")
+            norm = _memu_cat_name(raw)
+            if norm != raw and len(key_owners[_memu_cat_key(raw)]) == 1:
+                repairs.append((cat, norm))
+
+        # Occupancy must be judged in the domain the REBUILD keys on -- cat.name.lower()
+        # at the name-to-id rebuild in _initialize_impl, which does NOT strip -- and
+        # over the names the rows will carry AFTER the repairs above.  A key that only
+        # exists as a _memu_cat_key of some padded row is not one any consumer computes,
+        # so treating it as taken suppresses the seed that would supply it.
+        renamed = {id(row): new for row, new in repairs}
+        existing = {renamed.get(id(cat), getattr(cat, "name", "")).lower() for cat in rows}
         # Snapshot: never iterate the advertised list while seeding from it.
-        missing = [c for c in list(self._service.category_configs) if c.name not in existing]
-        await self._seed_categories(missing)
+        missing = [
+            c for c in list(self._service.category_configs)
+            if _memu_cat_name(c.name).lower() not in existing
+        ]
+        await self._seed_categories(missing, repairs)
 
-    async def _seed_categories(self, cat_cfgs: list[Any]) -> None:
-        """Persist startup seed categories WITHOUT touching the advertised set.
+    async def _seed_categories(
+        self, cat_cfgs: list[Any], repairs: list[tuple[Any, str]] | None = None,
+    ) -> None:
+        """Write the planned repairs and seeds, embedding once before any write.
 
-        memU built ``category_configs`` / ``category_config_map`` /
-        ``_category_prompt_str`` from these same entries before this runs, so
-        appending to them here (what ``_create_category_impl`` does for runtime
-        creation of a category memU does *not* yet advertise) would list every
-        category twice in the memorize prompt.
+        Seeds go through the repository rather than ``_create_category_impl``,
+        which also appends a ``CategoryConfig``: memU built ``category_configs`` /
+        ``category_config_map`` / ``_category_prompt_str`` from those entries
+        before this runs, so appending here would advertise every category twice.
         """
-        if not self._service or not cat_cfgs:
+        repairs = list(repairs or ())
+        if not self._service or (not cat_cfgs and not repairs):
             return
 
-        # Embed before any write.  A row persisted with a null embedding is never
-        # repaired by a later boot (get_or_create_category returns an existing row
-        # untouched) and both category rankers skip null vectors, so an embed
-        # failure must propagate instead of falling back to None.  One batched
-        # call, matching memU's own category initializer.
-        embeddings: list[Any] = [None] * len(cat_cfgs)
+        # Embed before any write, in ONE batched call covering repairs and seeds
+        # alike.  A row persisted with a null embedding is never repaired by a
+        # later boot (get_or_create_category returns an existing row untouched)
+        # and both rankers skip null vectors, so a failure must propagate rather
+        # than fall back to None -- and it must not leave a partial migration behind.
+        # A repaired row is re-embedded from its NORMALIZED text for the same
+        # reason its name is normalized: the stored vector is what cosine_topk
+        # ranks, so seeds and repairs have to share one space.
+        plan: list[tuple[str, Any]] = (
+            [("repair", r) for r in repairs] + [("seed", c) for c in cat_cfgs]
+        )
+        texts: list[str] = []
+        for kind, item in plan:
+            src = item[0] if kind == "repair" else item
+            texts.append(_memu_cat_embed_text(
+                getattr(src, "name", ""), getattr(src, "description", "") or "",
+            ))
+        embeddings: list[Any] = [None] * len(plan)
         if self._has_embeddings:
-            texts = [
-                f"{c.name}: {c.description}" if c.description else c.name
-                for c in cat_cfgs
-            ]
             embeddings = list(await self._service._get_llm_client("embedding").embed(texts))
 
         # Pair up front, so a provider returning the wrong number of vectors fails
         # before the first write rather than part-way through the loop.
-        pairs = list(zip(cat_cfgs, embeddings, strict=True))
+        pairs = list(zip(plan, embeddings, strict=True))
 
         repo = self._service.database.memory_category_repo
-        for cat_cfg, embedding in pairs:
+        for (kind, item), embedding in pairs:
+            if kind == "repair":
+                row, name = item
+                repo.update_category(category_id=row.id, name=name, embedding=embedding)
+                logger.info("Repaired category name: %r -> %s", getattr(row, "name", ""), name)
+                await self._audit(
+                    "category_updated", "category", row.id, "bridge",
+                    {"name_before": getattr(row, "name", ""), "name_after": name},
+                )
+                continue
+            # Store memU's normalized form: it is the name the memorize prompt
+            # shows the LLM and the only one its reverse lookups can resolve.
+            name = _memu_cat_name(item.name)
+            description = item.description.strip()
             repo.get_or_create_category(
-                name=cat_cfg.name,
-                description=cat_cfg.description,
+                name=name,
+                description=description,
                 embedding=embedding,
                 user_data={},
             )
-            logger.info("Seeded category: %s", cat_cfg.name)
+            # Audit the STORED name, so target_id identifies what was written.
+            logger.info("Seeded category: %s", name)
             await self._audit(
-                "category_created", "category", cat_cfg.name, "bridge",
-                {"description": cat_cfg.description},
+                "category_created", "category", name, "bridge",
+                {"description": description},
             )
 
     # Maximum time (seconds) for a single memorize operation before cancellation.
--- a/nerve/memory/memu_bridge.py
+++ b/nerve/memory/memu_bridge.py
@@ -1912,8 +1912,12 @@ class MemUBridge:
             ]
             embeddings = list(await self._service._get_llm_client("embedding").embed(texts))
 
+        # Pair up front, so a provider returning the wrong number of vectors fails
+        # before the first write rather than part-way through the loop.
+        pairs = list(zip(cat_cfgs, embeddings, strict=True))
+
         repo = self._service.database.memory_category_repo
-        for cat_cfg, embedding in zip(cat_cfgs, embeddings, strict=True):
+        for cat_cfg, embedding in pairs:
             repo.get_or_create_category(
                 name=cat_cfg.name,
                 description=cat_cfg.description,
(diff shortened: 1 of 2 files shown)


oranjeai:
<details>
<summary>Internal second-model review -- 7 rounds, 37 findings (27 agreed and fixed, 6 disagreed with evidence, 4 dropped)</summary>

Before opening this PR I ran it through an independent review harness: each round a second
model reviewed the diff cold against the PR body, and a reviewer that never edits source
adjudicated every finding. Six rounds bounced back for fixes; this is the round that passed.
The point of publishing the ledger is that a maintainer can see what was challenged, what I
accepted, and -- more usefully -- where I was wrong and had to reverse myself.

**Six rounds each found a real defect in the previous round's fix.** That is the honest
summary: the original filed defect was round 1, and every round after it was a defect in my
own remedy. Worth knowing before reading the diff, because it explains why the change is
larger than "seed from the other list".

| # | Finding | Verdict |
|---|---|---|
| 1 | `init-probe-network` -- the new tests made 12 real outbound calls; artifacts claimed offline | ❌ agreed, fixed |
| 2 | `category-name-normalization` -- nerve stored a name memU's resolver could never look up | ❌ agreed, fixed (**reversed**, see below) |
| 3 | `category-key-invariant` -- normalizing only the comparison made two cells **worse than base** | ❌ agreed, fixed |
| 4 | `rename-claim-false` -- my claim "`update_category` cannot rename" was false in all three backends | ❌ agreed, fixed |
| 5 | `lookup-key-collision` -- the repair guard compared in the wrong key domain; two cells worse than base | ❌ agreed, fixed |
| 6 | `repair-embedding-order` -- a repaired row kept a vector embedded from its raw text, contradicting this PR's own seeding rule | ❌ agreed, fixed |
| 7 | `multi-owner-key-unaddressable` -- the key-domain move itself regressed against base when two rows share a key | ❌ agreed, fixed |
| 8 | `availability-late-failure-test` -- the `_available` move was pinned only by a failure raised *before* both publication points | ⚠️ agreed, fixed |
| 9 | `runtime-category-restart-test` -- the restart arm re-implemented the rebuild instead of executing it | ⚠️ agreed, fixed |
| 10 | `repair-audit-gap`, `seed-comment-inverted`, `rebuild-citations-stale`, `category-comment-overload`, `inert-regression-arms`, `conflict-census-incomplete`, `validation-count-stale` (+ 11 more) | 💡 agreed, fixed |

**Where I was wrong, in public:**

- `category-name-normalization` was raised as a major in round 1 and I **disagreed**. It came
  back as a blocker in round 2 with a citation I had not had -- memU shows the LLM
  `name.strip() or "Untitled"` -- and I **reversed**: the name the model is told to use was not
  the name nerve persisted or mapped. One of my three original grounds was measurably wrong
  (I claimed a `strip()` here would collide with a sibling branch; the function it lives in is
  introduced by this PR, so its interior is uncontested).
- Round 3's evidence argued a legacy un-resolvable row was "not a regression". It compared the
  wrong cell pair -- with the *config* padded, base is equally broken; with the config
  normalized (the real post-upgrade shape) base seeds a resolvable row and my code did not.
  That refutation is what surfaced the strictly better remedy.
- Rounds 5 and 6 each found the round before had introduced a fresh regression against base.
  Both were caught by measuring a store/config grid on all three trees rather than by reading.

**Disagreed, with evidence** (recorded rather than silently dropped):

- ⚠️ `unavailable-recall-empty` (blocker, this round). The mechanism is **real and I confirmed
  it**: `memory_recall_handler` is the only one of nine memory handlers that does not gate on
  `memory_bridge.available`, so an unavailable bridge is rendered to the agent as "No relevant
  memories found." rather than as an outage. It is not this PR's, on three counts:
  `nerve/agent/tools/handlers/memory.py` is **byte-identical** to the merge-base and the
  missing gate is present there at the same line; the muted state was already reachable at base
  from eight pre-publication failure sites, and this PR's `_available` move makes matters
  *better* (base published availability early, so a late failure advertised memory as usable on
  a half-initialized service -- this PR routes those failures into the state where eight of nine
  handlers say so explicitly). Decisively, the remedy the reviewer asked for is **already
  implemented on a sibling branch**, both halves: `recall()`'s guard raises
  `MemoryBackendUnavailable` instead of returning `[]`, and
  `test_t13_uninitialized_bridge_renders_unavailable` asserts through the real handler that the
  response says "MEMORY RECALL UNAVAILABLE" and does *not* say "No relevant memories found."
- ⚠️ `embed-init-timeout` (raised three times). The unbounded startup embed is real and
  unbounded, but this branch does not create or worsen it: on the configured path base ran one
  unbounded embed *per category* inside `_create_category_impl`, so one batched call is
  strictly fewer, and the only alternative producer (memU's own initializer) embeds unbounded
  too. It is owned by a separate PR that bounds all three call sites, which is the right place
  -- bounding is a property of the call sites, not of who calls them.
- ⚠️ `runtime-normalization-gap` (blocker). Real, and byte-identical to base:
  `_create_category_impl` is unchanged here and `nerve/gateway/` is untouched. Owned by open
  PR #251, which normalizes exactly those sites.
- 💡 `base-passing-tests`, `startup-embed-timeout` -- refuted by mutation rather than argument:
  each named arm is killed by at least one per-site mutant, and the three that pass on both
  trees are deliberate no-regression guards, labelled as such.

**Coordination.** This branch conflicts in both files with open PRs #251 and #249, and #251 is
the semantic collision: it fixes the same double-advertising problem by keeping the seed loop
and avoiding the re-append, where this PR replaces the loop. Whoever lands second should keep
exactly one of the two mechanisms, not merge both. #248 and #70 overlap mechanically only.
Nothing was rebased, resolved or stacked.

Review cost for this PR: **$90.28** across 13 gated runs (6 approach rounds, 7 diff rounds).
Two of the seven diff reviews completed but lost their ledger row to a harness reap; their
findings were recovered from the surviving raw response and used only to drive fix rounds,
never to authorize publication -- this round's pass row is a real one.

</details>

oranjeai:
<details>

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes, 100% not a rate. Empty `memory.categories` + empty DB, one process: `advertised_to_LLM = 10`, `_ensure_categories body ran = False`, `map_size = 0`, `db_rows = 0`, `resolvable = 0 of 10`. Second, independent defect on the **configured** path (cold store): 3 configured categories give `advertised 3 -> 6, duplicated = 3, prompt_lines = 6`. |
| b | **Root cause explained?** | memU's `MemorizeConfig` default factory supplies 10 `CategoryConfig`s and `MemoryService.__init__` formats `category_configs` into the memorize prompt, so those 10 are what the LLM is told. nerve forwards `memory_categories` only when its own config is non-empty, sets `categories_ready = True` (suppressing memU's initializer), and `_ensure_categories` early-returned on an empty config. The set that was advertised and the set that was seeded were different objects. The name-to-id rebuild could not compensate: it reads the in-memory repo cache, and the only `list_categories()` warming it sat behind that early return. Unmapped names are dropped with no error. |
| c | **Fix matches root cause?** | Yes: seed from the effective advertised set through a non-advertising repository write, load persisted rows unconditionally, embed before any write, publish availability last. No widened bound, no defensive guard, no suppression. |
| d | **Test intent preserved / new tests added?** | 44 new cases. Count agrees two independent ways: `--collect-only` on the file 76 -> 120 (sorted NAME-set delta +44 added, 0 removed), and suite passed 2933 -> 2977, both +44. One existing arm was *replaced* rather than removed: `test_a_legacy_raw_row_is_left_alone_not_duplicated` became `test_a_legacy_raw_row_is_repaired_not_duplicated`, because the fix changed that behaviour from recognise-and-leave to repair-by-rename. No existing test removed or weakened. |
| e | **Both directions demonstrated?** | Yes. Base source `94406ea` with the final tests kept: **41 failed / 79 passed** at file scope; with the fix all 118 collect and only the 6 pre-existing timezone artifacts fail, the same 6 that fail on unmodified `origin/main`. Against the tree before the normalization round (`5efb96d`): **5 failed / 18 passed**, i.e. exactly the 5 arms that round adds, the end-to-end one failing on the resolution mismatch (`assert 0 == 2`) rather than any incidental error. Every base-arm failure was read and lands on the intended assertion (`assert [] == ['cat_0'...]`, `assert 'work' in {}`, `DID NOT RAISE`, `0 == 1`, `True is False`). The 3 base-passing cases are pinned non-vacuous by mutation. **12 mutants, all killed**, each asserting its anchor applied exactly once, with an unmodified control at both ends of the matrix. A later fix round added 3 more mutants for the probes' offline guard (warmup stub removed, socket blocker removed, and a semantically identical no-op), all resolved as expected. The normalization round adds **6 more** (one per normalization site, one identity-helper, one no-op): all resolved as expected, control green at both ends, `TREE_RESTORED_OK`. The per-site mutant for the already-exists comparison first **survived**, which was not a harness artifact: it exposed an uncovered reachable state (a store already holding a raw-named row). Review then measured that normalizing only the comparison made that state *worse than base* -- the seed base performed is suppressed while the row stays unresolvable -- so the fix now **renames** the legacy row instead. The r3 matrix is **9 mutants, all resolved as expected** (control 27 passed at both ends, `TREE_RESTORED_OK`), including three for the repair itself: the `repo.update_category(...)` call deleted, a no-op rename to the raw name, and the repair guard keyed on the rows walked so far instead of the whole set -- that last one killed by exactly one arm, the test written for a real order-dependence defect found while implementing (`list_categories()` applies no `ORDER BY`, so the prescribed guard produced two identically-named rows when the raw row came first). The r4 matrix is **15 mutants, all resolved as expected** (control 37 passed at both ends, `TREE_RESTORED_OK`): 6 new -- occupancy keyed on the exact name, the single-owner test dropped, the lookup-key helper not lowercasing, the repair written before the embed batch, repairs embedded from their raw name, and the repair audit removed -- plus the whole r3 set **re-expressed against the amended source**, each preserving the defect it modelled rather than its literal text, because a surviving inherited mutant would be a coverage regression this round caused. None survived. Because this round fixes defects of two generations, both directions are shown per generation on clean `git archive` exports: against `origin/main` **41 failed / 79 passed**, against the previous commit `b475ffd` **19 failed / 99 passed**, and on the amended tree **6 failed / 112 passed** -- figures re-measured with the final test set, in which that round's 8 discriminating arms account for 8 of the `b475ffd` failures alongside the 6 artifacts. **3 of the 10 new arms fail only at `b475ffd`** -- the key-collision arms, whose defect this branch itself introduced -- so a single base arm could not have discriminated them. Two more pass in both arms by construction and are labelled preservation guards / behaviour pins rather than reported as coverage. A fifth round then found that the previous round's key-domain move had itself introduced a regression against base: with two or more stored rows sharing one lookup key and none of them already normalized, occupancy was judged in a domain the name-to-id rebuild does not use, so the advertised name was marked present, no row was seeded, and nothing resolved -- while base seeds a third row and does resolve. Occupancy now uses the rebuild's own key over post-repair names. Its **5 new arms** (two parametrized pairs over both insertion orders, plus an end-to-end upgrade) fail at BOTH earlier generations and, deliberately measured, for **different reasons**: at the previous commit `resolved_prompt` is empty (`0 == 1`), while at `origin/main` resolution succeeds and the arm fails on the row set instead -- which is the measurement proving the defect was a regression rather than an unfixed gap. Whole-file per generation: `origin/main` **41 failed / 79 passed**, previous commit **11 failed / 107 passed**, amended **6 failed / 114 passed**. The r5 matrix is **18 mutants, all resolved as expected** (control 42 passed at both ends, `TREE_RESTORED_OK`): 3 new -- occupancy reverted to the pre-repair key set, occupancy over post-repair names but stripped, and occupancy in the rebuild's domain but over pre-repair names -- plus the whole inherited set. The first is killed by the new arms in both insertion orders; the third is killed only by pre-existing repair arms, reported rather than smoothed over, since it shows the two halves of the fix are pinned by different arm sets. The matrix also asserts all three r5 arms are present in the tests it runs, so a survival cannot be a vacuous-export artifact. The 18-cell store/advertised census was re-run against the REAL amended source rather than the prototype model: the 14 pre-existing cells are byte-identical in rows, row count, key count and unresolved set, the 4 regressing cells all go to 0 unresolved, and no cell is left unresolvable. A sixth round adds the two arms that pin properties the PR body publishes but no arm could observe. `_available`'s move to the end of `_initialize_impl` was pinned only by a failure injected INSIDE `_ensure_categories`, which raises before both the old and the new publication point, so that arm passed at either position; a new `fail-late` probe mode raises at the interceptor registration instead -- after seeding, before publication, and outside the swallowing `try` -- and asserts seeding had already succeeded, which is what distinguishes it from the pre-existing arm. Its mutant moves the three publication statements back to immediately after the seeding call and kills **exactly one arm, the new one**, while the pre-existing arm still passes: that asymmetry is the proof, and the first version of the mutant (which DELETED rather than moved) was rejected because it killed three arms for the wrong reason. The restart property was covered only through a test-local re-implementation of the name-to-ID rebuild, so no arm observed the production rebuild; every full-initialize arm that pre-stored a row also configured the same name. A new full-initialize arm pre-stores an unadvertised row and asserts through the real `ctx.category_name_to_id`, failing at base on exactly `assert 'work' in {}`. The r6 matrix is **21 mutants, all resolved as expected** (control 44 passed at both ends, `TREE_RESTORED_OK`): the inherited 18 carried in byte-verbatim, every anchor still matching once, plus the availability move, the rebuild loop body deleted, and the base shape restored (early return plus the removed preload `list_categories()`). That last kills **29** arms -- reported rather than normalised, because restoring base's shape restores the whole defect, which also shows the arm pins the unconditional load and not merely the rebuild. |
| f | **Fix is general across code paths?** | Two category-creation paths exist in `nerve/`. Startup seeding is fixed; runtime create (`_create_category_impl`, reached from `/api/memory`) is **byte-identical to base** -- whole-method `diff` vs `origin/main` is empty -- because it must advertise, which is exactly why seeding must not use it. All 14 references to the advertised set and the map live in one module; `xmemory_bridge.py` has 0 category references. The normalization round is confined to `_ensure_categories` / `_seed_categories`. Occupancy is now judged in ONE domain everywhere it is judged -- the lookup key `name.strip().lower()` that the rebuild and `memu/app/memorize.py:682` both compute -- via a single `_memu_cat_key` helper, and the embedding text via a single `_memu_cat_embed_text` that is memU's `_category_embedding_text` in one place instead of inline, so seeds and repairs provably share one text function. The name-to-id rebuild and `_create_category_impl` are deliberately left raw and stay byte-identical to base: once the row is stored normalized the rebuild's `cat.name.lower()` already yields memU's lookup key, and both sites are contested with the sibling branch (see Coordination). Runtime path re-verified alongside the fix: after init (10/10/10), creating a category gives advertised 11, 0 duplicates, rows 11, map 11, invariant 11 of 11. |
| g | **Fix generalizes across inputs (params/datatypes/wrappers)?** | Probed through the real store: 0 categories, 1 category, empty description, case-differing names, duplicate advertised name, unicode/CJK, 300-char name + 900-char description, `/`/`:`/space in names, and embed returning too few / too many vectors. **This found a real gap:** the original lazy `zip(strict=True)` wrote 1 row before raising on a length mismatch. Fixed by materializing the pairs before the write loop -- now 0 rows -- pinned by two parametrized cases and mutant M12. The normalization round re-ran the same shape over the name domain -- padded, whitespace-only, empty, already-normalized, and a stored row that is itself padded -- and every differing case was enumerated before a fixture was written rather than sampled. The r4 round adds the class those fixtures structurally could not reach: padding COMBINED with a case difference. Measured on all three trees over a 4-cell table and a 36-cell grid (6 stored x 3 configured x 2 config modes), reporting `rows - keys` per cell because a bare key count is not comparable across trees that legitimately end with different row counts. |
| h | **Backward compatible?** | No setting default, serialization or format change. `get_or_create_category` is idempotent by name, so existing stores are untouched (same row ids across two runs). Two deliberate behaviour changes, both disclosed in the PR body: a default-config install now creates 10 rows on first boot, and an embedding outage on a provider-configured install now fails init loudly instead of persisting permanently-unrankable rows. |
| i | **Invariants and contracts preserved?** | Advertised set left exactly as memU built it (10 before / 10 after, 0 duplicates, one prompt line per category, both paths). Nothing half-written: embed precedes every write and pairing precedes the loop, so both failure modes leave the store empty -- asserted as the **absence** of rows, not merely a raised error. Availability now means fully initialized: implemented as a **move**, not a rollback, because a rollback in the generic handler would miss the dedicated `except ImportError` reachable via two later imports; verified no statement in the moved range reads `_available` and none of the five methods called there gates on it. Audit contract kept (`category_created` / `source="bridge"` per seed on both paths, zero on an idempotent re-init, no new audit source); the audit `target_id` and the log line now name the **stored** row rather than the raw config text, so they identify what was written. A row persisted earlier under a raw name is now **repaired**: it is renamed to memU's form so its rebuild key becomes the one every consumer computes. `nerve`'s own `update_category` wrapper cannot rename (it forwards summary/description only, and `_update_category_impl` pins `new_name = cat.name`), but the repo layer does, in all three backends, and the **row id survives**, so linked `category_items` are not orphaned -- asserted in three arms and end to end. Measured over a 24-cell grid (4 stored x 3 configured x 2 config modes): cells resolving every advertised name go base **12/24** -> **24/24**, with no cell worse than base. A raw row is repaired only when the rename can neither duplicate a name NOR land on a lookup key another row already owns: `'  Alpha  '` normalizes to a display name no row holds while sharing the key of a stored `'alpha'`, so renaming it would leave two live rows behind ONE `category_name_to_id` entry with the winner decided by cache order. Such a key is left entirely alone, so base's rows survive and no items are discarded. Measured: `rows - keys` per cell is at most **1** at base and at the previous commit, and **0** on this tree; over the 36-cell grid, **0** cells collapse more than base and **0** rows are unaddressable, while cells resolving every advertised name are **36/36** against base's **20/36**. Case-only pairs (`Alpha` / `alpha`) stay out of scope -- both are already normalized, base produces the same two rows, and merging them would have to discard one row's items -- pinned by its own arm. Repairs and seeds are planned first, embedded in ONE batched call, and only then written, so an embedding outage can no longer leave a renamed row beside an unseeded category; a repaired row is re-embedded from its normalized text, because both rankers read the stored vector and a name-only rename would leave that one row ranked in the space its raw text occupied. The rename is audited as `category_updated` against the row id, matching `_update_category_impl`, so this file has no unaudited category mutation. A repaired row keeps its own description: it may have been edited through the API, and the rename is not a reconciliation point -- pinned by an arm so a later widening has to change it deliberately. |

</details>

Session id: cron:clickhouse-impl-slot-4:20260803-221500

oranjeai:
@pufit could you take a look? You own most of `nerve/memory/memu_bridge.py` (10 of the last 20
commits there, including `cfc236f`), and this touches `_ensure_categories` plus the availability
publication point in `_initialize_impl`.

The two things worth your judgment rather than my measurement:

1. **Repairing persisted rows.** A row stored under a non-normalized name (`"  work  "`) is renamed
   in place so its lookup key becomes the one memU computes. It keeps its id, so linked
   `category_items` survive, and it is skipped entirely when another row already owns the target
   lookup key. That is a write to existing user data on upgrade, which felt like the right call over
   leaving permanently unresolvable rows behind, but it is a call.
2. **Failing init loudly.** An embedding outage or a category-table read error now leaves memory
   unavailable instead of proceeding with a half-built map. Base swallowed both. Since
   `engine.py:459` discards `initialize()`'s return, `_available` is the only signal the agent gets.

Note it conflicts with #251 in both files, and that conflict is semantic, not textual: #251 keeps the
`_create_category_impl` seed loop and avoids the re-append, this replaces the loop. Whichever lands
second should keep exactly one of the two mechanisms. #249 also conflicts (it adds a second consumer
of the rows seeded here); #248 and #70 overlap mechanically only.

oranjeai:

PR #260, opened 2026-08-04
memory: never report a failed recall as "no relevant memories"

oranjeai:
## Symptom

A `memory_recall` whose embedding call fails at the **connection layer** — DNS, refused,
reset, or a transport closed under it by a client reset — renders as `Recalled 0 memories` /
`No relevant memories found.` A session concludes memory is empty, redoing work: the
"confident wrong answer" `MemoryBackendUnavailable` exists to prevent (`835d857`).

Not theoretical — `Connection error.` appears **42** times in this instance's 1.95 GB
`nerve.log` (2026-08-03 21:10 UTC), all memorize/chat-side and **0** on `memU recall failed`:
the class occurs routinely, `recall` has not drawn it yet.

## Root cause

`recall()` enumerated transient *signatures* instead of partitioning failure *kinds*. Its only
term, `_is_transient_llm_error`, keys on an HTTP **status** (429/5xx/auth), so a failure with
no HTTP response cannot match it by construction and fell through to `return []` beside logic
errors. `retrieve` has **0** `except` arms, so nothing normalizes it.

## The fix

`[]` from `recall()` now means "retrieval returned nothing", for every failure it can observe:

- a new `_is_llm_transport_failure` tests each SDK's `APIConnectionError` base class —
  `openai` and `anthropic` are disjoint hierarchies, `_BedrockLLMClient` raises the latter,
  and each `APITimeoutError` subclasses its own, covering timeouts;
- infrastructure failures raise `MemoryBackendUnavailable`; **everything else propagates** to
  the caller's pre-existing generic arm, not swallowed;
- the uninitialized-bridge guard raises instead of returning `[]`;
- the message drops its false cause ("transient proxy/auth error") for an outcome-only one —
  the same arm covers 429s, which *are* responses.

A genuine miss still returns `[]`, pinned by a test. All three callers already catch, and
`engine.py`'s pre-recall stops **freezing** a failed recall into session metadata, replayed
on every rebuild.

## Validation

**14 fail / 1 pass** at base (the 1 is the genuine-miss control) vs **15 pass** fixed; a
17-mutant matrix kills every mutant. Whole suite 2933 -> 2948, **identical** failed set.

<details>
<summary>Scope, and why <code>[]</code> is not globally honest</summary>

The widening is `recall()` only. The write and category paths keep their `return
[]`/`False`: their tool layers already render an explicit failure (`Failed to update memory
<id>`, `NOT saved - MEMORY BACKEND UNAVAILABLE`), so they produce no wrong *answer* — only
a **read** reporting emptiness lies. `memorize_file` already raises for 429/5xx.

`[]` is honest for everything nerve *observes*, not globally: memU absorbs a malformed LLM
ranking response *below* this arm (`memu/app/retrieve.py:1344` logs and returns an empty
list), so on the `method="llm"` path a garbled ranker reply still yields a legitimate-looking
empty recall and no exception ever reaches `recall()`. `memu-py` is pinned, so closing that
is a memU-layer change - out of scope, named here so the docstring is not read as the
stronger claim.

Raw `httpx` errors are deliberately untested: both SDKs wrap them
(`openai/_base_client.py:1683`, `anthropic/_base_client.py:2061`), and nerve sets
`client_backend="sdk"` at every profile, so `HTTPLLMClient` is never constructed — that
family has no carrier.

Mutation matrix: 17 mutants, all killed, unmutated control green at both ends, including
`M12` (residual returns empty - under-classification) and `M13`/`M15` (over-classification).
The 7 pre-existing whole-suite failures are timezone off-by-ones in
`TestResolveEventDatesSync` plus one in `test_telegram_sessions`, unchanged by this PR;
`ruff` finding set identical modulo line shifts.
</details>

### DIFF
--- a/nerve/agent/tools/handlers/memory.py
+++ b/nerve/agent/tools/handlers/memory.py
@@ -107,9 +107,9 @@ async def memory_recall_handler(ctx: ToolContext, args: dict) -> ToolResult:
     except MemoryBackendUnavailable as e:
         logger.warning("Memory recall: backend unavailable: %s", e)
         memu_block = (
-            "⚠️ MEMORY BACKEND DOWN (transient proxy/auth error) — recall is UNAVAILABLE right now, "
-            "NOT empty. Do NOT conclude 'no relevant memories'. Reconstruct context from other "
-            "sources (git history, your own notes) and alert the operator that memory is down."
+            "⚠️ MEMORY RECALL UNAVAILABLE — this recall did not complete, so this is NOT an "
+            "empty result. Do NOT conclude 'no relevant memories'. Reconstruct context from other "
+            "sources (git history, your own notes) and alert the operator that this recall failed."
         )
     except Exception as e:
         logger.error("Memory recall failed: %s", e)
--- a/nerve/memory/memu_bridge.py
+++ b/nerve/memory/memu_bridge.py
@@ -37,11 +37,15 @@ logger = logging.getLogger(__name__)
 
 
 class MemoryBackendUnavailable(RuntimeError):
-    """Raised when a memU operation fails because the LLM backend (the shared
-    proxy route) is transiently unavailable — HTTP 429/5xx or an auth blip
-    (``auth_unavailable`` / revoked OAuth) — rather than because the query
-    genuinely had no results. The tool layer surfaces this LOUDLY so a session
-    never mistakes "memory is DOWN" for "no relevant memories"."""
+    """Raised when a memU operation could not complete because the backend was
+    unavailable, rather than because the query genuinely had no results. The
+    tool layer surfaces this LOUDLY so a session never mistakes "memory is
+    DOWN" for "no relevant memories".
+
+    Deliberately described by OUTCOME, not by cause: the raising sites already
+    cover an HTTP 429/5xx or auth blip, an LLM-call timeout, a transport-level
+    failure with no HTTP response, SQLite write-lock contention, and a bridge
+    that never initialized. Enumerating causes here goes stale."""
 
 
 def _is_transient_llm_error(exc: BaseException) -> bool:
@@ -70,6 +74,34 @@ def _is_transient_llm_error(exc: BaseException) -> bool:
     # Belt-and-suspenders: the proxy's stuck-cooldown signature regardless of
     # how the status surfaced.
     return "auth_unavailable" in low
+
+
+def _is_llm_transport_failure(exc: BaseException) -> bool:
+    """True if ``exc`` is a transport-level LLM failure: the call failed at the
+    connection layer rather than with an HTTP status (DNS, refused, reset,
+    closed transport; possibly mid-response).
+
+    Disjoint from :func:`_is_transient_llm_error`, which keys on an HTTP
+    *status*. Both mean the call produced no usable result, so recall must not
+    report "no results".
+
+    Both SDK families are needed and neither subsumes the other: memU's own
+    client raises ``openai.APIConnectionError``, ``_BedrockLLMClient`` raises
+    ``anthropic.APIConnectionError``. Imports are guarded so the module stays
+    importable without a given provider. Each SDK's ``APITimeoutError``
+    subclasses its ``APIConnectionError``, so the base-class test covers
+    timeouts; raw ``httpx`` errors need no arm because both SDKs wrap them."""
+    for _mod, _name in (
+        ("openai", "APIConnectionError"),
+        ("anthropic", "APIConnectionError"),
+    ):
+        try:
+            _exc_type = getattr(__import__(_mod), _name)
+        except (ImportError, AttributeError):
+            continue
+        if isinstance(exc, _exc_type):
+            return True
+    return False
 
 
 def _is_sqlite_locked_error(exc: BaseException) -> bool:
@@ -2786,9 +2818,14 @@ class MemUBridge:
         Args:
             limit: max item hits to return (full content).
             category_limit: max category breadcrumbs to return.
+
+        Raises:
+            MemoryBackendUnavailable: the backend could not be reached, or the
+                bridge never initialized. Any other exception propagates.
+                An empty list therefore means retrieval returned nothing.
         """
         if not self._available or not self._service:
-            return []
+            raise MemoryBackendUnavailable("memory backend is not initialized")
         op_id = self._metrics.begin_op("recall", query[:80])
         try:
             # Retrieve runs vector search (numpy) and possibly SQLite reads
@@ -2846,12 +2883,13 @@ class MemUBridge:
         except Exception as e:
             logger.error("memU recall failed: %s", e)
             self._metrics.end_op(op_id, success=False, error=str(e))
-            # A transient backend outage must not masquerade as "no results" —
-            # surface it so the caller can flag amnesia instead of trusting the
-            # empty list.
-            if _is_transient_llm_error(e):
+            # Infrastructure failures get the dedicated type; anything else
+            # propagates instead of being swallowed. (memU can still absorb a
+            # malformed ranking reply below this level and return a real `[]`.)
+            if (isinstance(e, TimeoutError) or _is_transient_llm_error(e)
+                    or _is_llm_transport_failure(e)):
                 raise MemoryBackendUnavailable(f"memory backend unavailable: {e}") from e
-            return []
+            raise
 
     async def expand_category(
         self, category_id: str, query: str = "", limit: int = 20,
--- a/tests/test_memu_bridge.py
+++ b/tests/test_memu_bridge.py
@@ -7,9 +7,13 @@ from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
 
+import httpx
+import openai
 import pytest
 import pytest_asyncio
 
+from nerve.agent.tools.handlers.memory import memory_recall_handler
+from nerve.agent.tools.registry import ToolContext
 from nerve.config import MemoryConfig, NerveConfig
 from nerve.memory.memu_bridge import (
     MemoryBackendUnavailable,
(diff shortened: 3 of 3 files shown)


oranjeai:
<details>
<summary>Internal second-model review (3 rounds, 20 findings, all adjudicated)</summary>

Before opening this PR I ran my own cold code review plus an independent second-model
review (codex) against the resulting code, over **3 rounds**. Every finding is listed with
its verdict, including the ones I disagreed with and why. 0 verdict reversals; no finding
was raised twice.

### Fixed in this PR

| # | finding | what was wrong | how it was proved |
|---|---|---|---|
| 1 | `engine-recall-test` | The pre-recall regression test rebuilt `engine.py`'s block as local variables, so the production caller and its real `update_session_metadata` write never ran; the assertion was about the test's own dict. | Rewritten to drive the real `_get_or_create_client` and read the session row back from the DB. A mutant that makes the engine's arm freeze the failed recall **survives** the old test tree and **dies** against the new one. |
| 2 | `engine-recall-attempt` | Even after that repair the test still passed if pre-recall was skipped entirely: "client created" and "metadata absent" are also the outcomes when the branch never executes. | A mutant gating the `elif` to `False` **survived at 15 passed / 0 failed**. Fixed by asserting the injected mock was awaited; that mutant now kills exactly that arm. |
| 3 | `genuine-miss-attempt` | The genuine-miss control, the arm that keeps the re-raise inversion honest, could not distinguish a successful empty retrieval from a skipped one: its only assertion holds for any early `[]`. | A mutant returning `[]` before retrieval killed 12 arms but **left the miss control passing**. Fixed with an invocation assertion; that mutant now kills it. |
| 4 | `outcome-message-down` | The message is meant to describe the **outcome**, not the cause, and its own test rejects "BACKEND DOWN" as a false claim, yet it still ended "alert the operator that memory is down" -- false for two cases the same arm covers (a locally self-closed transport, and a bridge that never initialized). The case-sensitive oracle could not see the lowercase form. | Reworded to an outcome statement, and the missing needle added to the false-claim oracle so it covers the class it claims to. Reverting the reword now fails that arm. |
| 5 | `recall-comment-volume` | New docstrings and comments carried PR rationale and dependency-internal line numbers into production source -- citations that rot silently on a dependency bump. | Condensed to the outcome invariant plus the dual-SDK rationale (-20/+12 lines); the scope adjudication and dependency references moved to the PR body. |
| 6 | `session-context-arm-unasserted` | The second production caller (`session_context`) had no test making its recall raise, so the claim that it surfaces the error was unverified. | An arm was added; a mutant that makes that handler render the pre-existing empty text now kills it. |
| 7 | `pr-body-figures-stale-r1` | A fix round changed five figures the PR description quoted (arm counts, suite total, matrix size), leaving the description describing an earlier tree. | Re-derived each from the round's own logs and corrected. |
| 8 | `ai-gate-figures-unreproducible` | Two of my own figures did not reproduce: the count of broad `except` arms had drifted to 52-53 across rounds, and one embedding call-site count was off. | Re-measured by AST: **50** at both base and head. Both load-bearing conclusions were independently re-verified and stand. |
| 9 | `pr-body-mutant-count-stale-r2` | The final round added three mutants, making the description's "14-mutant matrix" wrong in two places while another comment already said 17 -- the PR would have contradicted itself across two published surfaces. | Re-derived from the mutation log: **17** distinct mutants, all killed. Both stale figures understated coverage. Corrected before the final gate ran, so the gate reviewed the corrected description. |
| 10 | `embed-site-count-off-by-one-r2` | A claim placed five embedding call sites "in `retrieve`"; by AST that function contains none of them -- they live in the pipeline steps it dispatches. | Corrected to name the true locus with line numbers. Every other figure in that row reproduced exactly. |

### Disagreed, with evidence

| finding | why I disagreed |
|---|---|
| `engine-freeze-success-side-unpinned` (mine) | Nothing pins the **positive** direction of the pre-recall freeze -- that a *successful* recall still writes the key, the deliberate prompt-cache behaviour. A real gap, but the base had zero coverage of that key either, and this change's contract is failure classification. Named rather than silently widened. |
| `register-handler-not-unregistered-in-t5` (mine) | The new engine test leaks one entry into the module-global interactive-handler registry. It follows the repo's own existing precedent for this shape, the key is unique to the test, and the whole suite is green with a failed-name set identical to base. Not this change's convention to alter. |
| `stray-magicmock-dir-untracked` (mine) | Harness artifact: 0 files / 88 empty dirs, untracked, absent from the commit, produced by *running* pytest -- present in unrelated checkouts and absent from one that never ran the suite. |
| `openai-not-a-declared-dependency` (mine) | The new test-only import arrives transitively from the hard-pinned `memu-py`, which imports it unconditionally, so any install where these tests run has it. The production predicate keeps its guarded import, so no source coupling is added. |

### Noted, not blocking

**Test-docstring volume.** The new test class carries more narrative than this repository's
convention prefers: 47 docstring lines against 245 body lines (**19.2%**), where the
next-busiest class in the same file is 5.9% and 12 of 17 are at 0.0%. Three spots restate
the symptom-and-fix story, embed a production occurrence count, and enumerate every
rejected wording. The equivalent finding on the *production* source was fixed in an
earlier round; this is the same class one file over, and it is tracked as a follow-up
rather than bounced, since no blocking finding remained. The embedded count is a lower
bound and is still true, so the defect is invited future drift, not a false statement.

**Three coverage observations, all on lines this change does not touch**, each demonstrated
by a surviving mutant and each recorded rather than papered over: the guard's
service-object conjunct is unpinned (that state is close to unreachable, and the guard's
changed behaviour *is* pinned); the exception-cause chain is unpinned (every arm matches on
a message substring the mutant still satisfies); and the success-side freeze is unpinned as
noted above. Every direction of the behaviour this change *does* alter is pinned: all 15
arms are killed by at least one mutant, with no arm left without a killer.

**One residual is deliberately out of scope** and stated in the PR description rather than
papered over: memU can absorb a malformed ranking response *below* this arm, so an empty
recall is honest for everything this code observes, not globally.

</details>

oranjeai:
<details>

| # | check | answer |
|---|---|---|
| **a** | Deterministic repro? | **Yes.** Classification is a pure function of the exception object -- no timing, no randomization, not a "~1%" case. Two independent reproductions: a 15-input predicate matrix (`BAD=0`), and end-to-end through the real `recall()` + the real tool handler, where the base tree renders `No relevant memories found.` for `openai.APIConnectionError`. |
| **b** | Root cause explained? | **Yes.** The embedding transport fails without a response -> `openai` raises `APIConnectionError` after its own retries -> memU re-raises untouched (`memu/app/retrieve.py:42 retrieve` has **0** `except` arms, verified by AST) -> `_is_transient_llm_error` keys on an HTTP **status** and so cannot match by construction -> `return []` -> the tool layer prints a count. |
| **c** | Fix matches the root cause? | **Yes** -- it repairs the *partition* (a base-class test across both SDK connection-error hierarchies, plus the plain-`asyncio` timeout term) rather than appending a special case. Not a band-aid: the tempting narrow alternative (key on a `__cause__` message for `client has been closed`) is mutant **M11**, killed by 4 arms, because it fixes the **0-occurrence** sub-case and leaves the 42-occurrence one broken. |
| **d** | Test intent preserved / new tests added? | **Yes.** 15 new arms; nothing weakened or removed. Two drive the real production callers rather than a copy of their logic (the engine pre-recall path via `_get_or_create_client` plus the persisted session row, and `session_context`). T14 pins that a genuine miss still returns `[]` (else the fix would make every miss look like an outage); T9 pins the **pre-existing** 429 case so the message rewording did not break the case it already covered. |
| **e** | Demonstrated in BOTH directions? | **Yes.** Base source + new tests: **14 failed / 1 passed** (the 1 is the genuine-miss control, which must hold on both trees). Fixed: **15 passed**. Plus a **17**-mutant matrix, every mutant killed, each hitting exactly its predicted arm set, unmutated control green at **both** ends, `TREE_RESTORED_OK`. Five mutants target the real production callers or the operator message and were measured to **survive** an earlier test tree while being **killed** by the current one, so every added assertion's coverage is demonstrated, not asserted. |
| **f** | General across CODE paths? | **Yes.** Enumerated all **50** broad `except` arms by AST (bare/`Exception`/`BaseException`, same count at HEAD and at base `94406ea`). After the fix exactly two read paths still `return []` -- `list_items` and `list_categories` -- and both are non-network-bearing, verified mechanically rather than assumed: **within the two callee function bodies** (`memu/app/crud.py:38,59`) there are **0** `embed`/`llm`/`client` references (whole-file `crud.py` has many; the claim is function-scoped), against **5** `embed_client.embed` call sites on the recall path (in the pipeline steps `retrieve` dispatches, `memu/app/retrieve.py:271,321,357,397,421`), so a pure DB read cannot fail for transport reasons. Fixed at the source (the classifier), not guarded at the render site. |
| **g** | Generalizes across INPUTS? | **Yes**, `BAD=0` over three axes: subclass **leaves** of both SDK families (so the fix cannot rot as an SDK adds leaves); **10 `__cause__` shapes** (None, closed transport, refused, DNS, `RemoteProtocolError`, read/write/pool errors, `OSError`, `ValueError`) -- all classify, since the predicate reads the type, never a message; and **6 degenerate inputs** (`BaseException`, `KeyboardInterrupt`, `SystemExit`, bare `Exception`, `GeneratorExit`, and an exception whose `__str__` **raises**) -- all `False`, none raised. |
| **h** | Backward compatible? | **Yes** -- no setting, no serialization format, no config key, no feature gate, no DB migration. The only behaviour change is which of two **existing** tool messages an agent sees. Import safety verified in 5 arms: with `anthropic` absent the `openai` arm still works; with **both** absent the predicate returns `False` without raising; a renamed SDK attribute is caught; and no module-level SDK import is added. |
| **i** | Invariants and contracts preserved? | **Yes**, and the invariant is stated **scoped**: *no failure that reaches `MemUBridge.recall` is reported as `[]`* -- deliberately **not** the global claim "`[]` always means a successful miss", because memU absorbs a malformed LLM ranking response *below* this arm (`memu/app/retrieve.py:1341-1347`, `memu-py==1.4.0` pinned), so a garbled ranker reply still yields a legitimate empty recall with no exception reaching us. Closing that is a memU-layer change, out of scope, named so nobody is misled. Re-raise safety verified at **all three** callers (each already catches broadly), so the inversion cannot escape unhandled. A predicate bug must **surface**, not degrade to `[]`: the `try` wraps only the import, never the `isinstance` -- proven with a metaclass whose `__instancecheck__` raises, which propagates. |

**Deliberate scope exclusions** (decisions, not omissions): `memorize_file`, `update_item` and
the two category paths keep their current returns -- their tool layers already render an
explicit failure, so they produce no confident wrong *answer*; the two category paths are also
owned by an in-flight sibling review. `_is_llm_timeout` stays in the file with its
`memorize_file` call site intact -- only `recall`'s arm no longer calls it, since each SDK's
`APITimeoutError` is a subclass of its `APIConnectionError` (pinned by an arm, so a future
reparenting fails loudly).

**Predicted mutation survival, recorded rather than papered over:** reverting the
`MemoryBackendUnavailable` docstring leaves the suite green (15/15 since r1 added an arm).
A docstring has no runtime observer, so no test can kill it; asserting on docstring text would
be pinning prose, not behaviour.

**Not applicable:** no 50/50 randomized-run plan and no Build ID -- this is a Python change in
this repo, not a ClickHouse test.
</details>

oranjeai:

PR #261, opened 2026-08-04
Derive task_write's deadline column from the content it wrote

oranjeai:
### Description

`task_write` overwrites a task's markdown file and then re-syncs the DB row, but it took
the `deadline` from a snapshot read *before* the file was replaced. Content carrying no
`**Deadline:**` line kept the snapshot value, so the row asserted a deadline the file it
had just written denied.

**Root cause.** That column is a pure projection of the `**Deadline:**` line. It has
exactly one SQL writer (the `upsert_task` INSERT) and no row-only setter, unlike `status`
and `tags`, which each have a dedicated `update_task_*` method. So no row state exists to
preserve for it, only a file to agree with.

**Effects**, with a concurrent writer committing inside the handler's snapshot-to-upsert
window: a deadline it set was dropped, and one it deliberately *cleared* was resurrected.
`escalation.py` reads this column, so a stale row escalates on a date nothing asks for.

**The change** takes the value from the content, removing the only stale read. No new
sentinel, signature, SQL or schema change: the existing `deadline=new_deadline or None`
already writes NULL when the file states nothing.

**Observable change:** a `task_write` whose content omits `**Deadline:**` now clears the
column, so `escalation.py` stops reminding on that task until a `**Deadline:**` line comes
back. Nothing is lost destructively: the file is authoritative, and `TaskManager.reindex`
derives the same value from it. (That indexer has no production caller today; it is cited
as the file-to-row definition, not a live path.) No test pinned the old behaviour.

**Validation.** Base arm (source reverted, tests kept) fails the three intended
assertions; fixed arm passes 4. Full suite 7 failed / 2937 passed, failure name set
identical to the 7 pre-existing failures at this base. `ruff` clean.

**Not closed here:** the PATCH route in `routes/tasks.py` carries the same snapshot echo
on the same argument, and the file/DB pair is still not atomic.

<details>
<summary>Test design and mutation matrix</summary>

The regression test drives the real handler with a one-shot wrapper on `db.get_task` that
fires the rival writer inside the window, and asserts both that the hook fired and that
the rival's write committed before asserting any outcome. A rival landing before the
snapshot is read cannot expose the bug, so without those guards the two concurrency cases
degenerate into duplicates of the single-threaded case and pass vacuously, verified by
neutering the rival with both guards removed, which turns the suite green.

Mutation matrix, unmutated control at both ends of every run, each mutant asserting it
applied before the suite runs:

| mutant | verdict |
|---|---|
| revert the line to the snapshot fallback | KILLED (exactly the 3 intended tests) |
| drop the `""` default | SURVIVED, a proven no-op: one consumer, and `or None` maps `""` and `None` identically in every content shape (no line / set / empty / whitespace / no frontmatter) |
| delete the hook-fired assertion | survives on the fixed tree (it is a vacuity guard, not a discriminator); killed against the base tree, and the paired "rival committed" assertion still catches a neutered rival |

Four cases: rival moves the deadline, rival clears it, content states a deadline
(unchanged behaviour, passes before the fix), and the single-threaded observable change.
Each asserts the row against a real `TaskManager.reindex()` oracle rather than a
hand-computed expectation.

</details>

### DIFF
--- a/nerve/agent/tools/handlers/tasks.py
+++ b/nerve/agent/tools/handlers/tasks.py
@@ -429,7 +429,9 @@ async def task_write_handler(ctx: ToolContext, args: dict) -> ToolResult:
     )
     new_title = parse_task_title(new_content) or task["title"]
     frontmatter = parse_task_frontmatter(new_content)
-    new_deadline = frontmatter.get("deadline", task.get("deadline", ""))
+    # The deadline column is a pure projection of the file's Deadline line, so it
+    # comes from the content we just wrote, never from the pre-write snapshot.
+    new_deadline = frontmatter.get("deadline", "")
     new_tags = tags_to_string(parse_tags_string(frontmatter.get("tags", task.get("tags", ""))))
 
     await ctx.db.upsert_task(
--- a/tests/test_task_write_deadline.py
+++ b/tests/test_task_write_deadline.py
@@ -1,0 +1,180 @@
+"""``task_write`` must derive the ``deadline`` column from the content it wrote.
+
+``task_write`` overwrites a task's markdown file and then re-syncs the DB row.
+The ``deadline`` column is a pure projection of the file's ``**Deadline:**``
+line: it has exactly one SQL writer (the ``upsert_task`` INSERT) and no
+row-only setter, unlike ``status`` and ``tags`` which each have a dedicated
+``update_task_*`` method. So there is no row state to preserve for it -- only a
+file to agree with.
+
+The handler used to fall back to a snapshot taken *before* the file was
+replaced, which made the row assert a deadline the file it had just written
+denied: a concurrent writer's value was dropped, and a deliberately cleared
+deadline was resurrected. ``escalation.py`` reads that column, so a stale row
+escalates on a date nothing asks for.
+
+The rival writer here commits strictly *inside* the handler's snapshot ->
+upsert window, via a one-shot wrapper on ``db.get_task``. Each test asserts the
+hook fired and the rival's write was committed before asserting anything about
+the outcome -- a rival that lands before the snapshot is read cannot expose the
+bug at all, so without those assertions the tests would pass vacuously.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from nerve.agent.tools.handlers import tasks as task_handlers
+from nerve.agent.tools.registry import ToolContext
+from nerve.db import Database
+from nerve.tasks.manager import TaskManager
+from nerve.tasks.models import parse_task_frontmatter
+
+TASK_ID = "t1"
+REL_PATH = f"memory/tasks/active/{TASK_ID}.md"
+
+WITH_DEADLINE = "# T1\n\n**Tags:** urgent\n**Deadline:** 2030-01-01\n\nbody\n"
+NO_DEADLINE_LINE = "# T1\n\n**Tags:** urgent\n\nbody edited\n"
+STATES_DEADLINE = "# T1\n\n**Tags:** urgent\n**Deadline:** 2030-01-01\n\nbody edited\n"
+
+RIVAL_DEADLINE = "2030-06-30"
+RIVAL_CONTENT = WITH_DEADLINE.replace("2030-01-01", RIVAL_DEADLINE)
+CLEARED_CONTENT = WITH_DEADLINE.replace("**Deadline:** 2030-01-01\n", "")
+
+
+async def _seed(tmp_path, db, monkeypatch):
+    """Workspace + row holding ``2030-01-01``, ready for ``task_write``."""
+    workspace = tmp_path / "ws"
+    (workspace / "memory" / "tasks" / "active").mkdir(parents=True)
+    (workspace / "memory" / "tasks" / "done").mkdir(parents=True)
+    (workspace / REL_PATH).write_text(WITH_DEADLINE, encoding="utf-8")
+    await db.upsert_task(
+        task_id=TASK_ID, file_path=REL_PATH, title="T1", status="in_progress",
+        deadline="2030-01-01", tags="urgent", content=WITH_DEADLINE,
+    )
+    monkeypatch.setattr(task_handlers, "_tasks_read", {TASK_ID})
+    return workspace, ToolContext(session_id="test", db=db, workspace=workspace)
+
+
+def _arm_rival(db, workspace, *, clear: bool):
+    """Fire one rival write right after the handler reads its snapshot.
+
+    Returns a dict the caller must check: ``fired`` proves the hook ran (so the
+    rival really is inside the window) and ``committed`` proves the rival's
+    write reached the DB.
+    """
+    state: dict[str, object] = {"fired": False, "committed": "<never>"}
+    original_get_task = db.get_task
+
+    async def hooked(task_id):
+        row = await original_get_task(task_id)
+        if not state["fired"]:
+            state["fired"] = True
+            content = CLEARED_CONTENT if clear else RIVAL_CONTENT
+            (workspace / REL_PATH).write_text(content, encoding="utf-8")
+            await db.upsert_task(
+                task_id=TASK_ID, file_path=REL_PATH, title="T1",
+                status="in_progress", tags="urgent",
+                deadline=None if clear else RIVAL_DEADLINE, content=content,
+            )
+            state["committed"] = (await original_get_task(TASK_ID))["deadline"]
+        return row
+
+    db.get_task = hooked
+    return state
+
+
+def _assert_rival_landed(state, *, clear: bool):
+    assert state["fired"], "harness: the get_task hook never fired"
+    expected = None if clear else RIVAL_DEADLINE
+    assert state["committed"] == expected, (
+        f"harness: rival did not commit ({state['committed']!r} != {expected!r})"
+    )
+
+
+async def _reindex_oracle(tmp_path, db_path_name, workspace):
+    """What the indexer derives from the file, i.e. the authoritative answer."""
+    oracle_db = Database(tmp_path / db_path_name, workspace=workspace)
+    await oracle_db.connect()
+    try:
+        await TaskManager(workspace, oracle_db).reindex()
+        row = await oracle_db.get_task(TASK_ID)
+        return row["deadline"] if row else None
+    finally:
+        await oracle_db.close()
+
+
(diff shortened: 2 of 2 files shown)


oranjeai:
<details>

| # | Question | Answer |
|---|---|---|
| a | Deterministic repro? | Yes. A one-shot wrapper on `db.get_task` fires the rival writer strictly inside the handler's snapshot-to-upsert window and asserts it committed; 3/3 cases every run. 3 of the 4 committed tests also fail deterministically on the base tree. No statistical repro anywhere. |
| b | Root cause explained? | Yes. The `deadline` column is a pure projection of the file's `**Deadline:**` line: one SQL writer (the `upsert_task` INSERT) and no row-only setter, unlike `status`/`tags` which each have a dedicated `update_task_*` method. The handler mixed in a value read before it replaced the file, so the row asserted a deadline the file denied. |
| c | Fix matches root cause? | Yes, and not a band-aid: it deletes the only read whose staleness mattered instead of guarding the symptom. No defensive check, lock or retry. Two alternatives were measured and are worse: a keep-sentinel still writes the rival's value into a row whose file has no line (1/8 wrong vs 0/8) and costs a new SQL sentinel across 7 call sites; routing through a partial-field update never writes `file_path`. |
| d | Test intent preserved / new tests added? | Yes. No existing test weakened, disabled or bound-widened. New `tests/test_task_write_deadline.py` with 4 cases, one of which exists solely to pin unchanged behaviour. No existing test mentions `task_write` and `deadline` together, so nothing pinned the old behaviour. |
| e | Both directions? | Yes. Base arm reverts **only** the source file and keeps the tests, so the tests are the discriminator: 3 failed / 1 passed, each failure traced to its intended `row["deadline"] is None` assertion line, with 0 harness assertions fired. Fixed arm: 4 passed. |
| f | General across code paths? | Yes, swept. Every snapshot echo on a projection argument was enumerated (8 sites) and dispositioned: the handler is fixed; `reindex` is already file-derived and is now what the handler agrees with; `task_update` rewrites the file's line in the same breath so row and file stay in agreement; `task_done`/`mark_done` copy the file verbatim; `task_create` writes the line it indexes. The PATCH route in `routes/tasks.py` is genuinely the same class and is left out on ownership, disclosed in the description as not closed. `source`/`source_url` share the None-means-NULL property but no file line expresses them, so they are not projections. |
| g | Generalizes across inputs? | Yes. The change is at a `dict.get` default, so the domain is content shapes, and every one was measured: no `**Deadline:**` line, value set, empty value, whitespace-only value, no frontmatter at all. The value reaching the consumer is identical with and without the `""` default in all five. Rival matrix in the tests: moves / clears / absent, against content that states or omits the line. |
| h | Backward compatible? | No schema change, no migration, no signature or SQL change, no settings default. One observable behaviour change, disclosed in the description: content omitting `**Deadline:**` now clears the column. Non-destructive and recoverable, since the file is authoritative and a reindex of it yields the same result, which the tests assert against a real reindex oracle. |
| i | Invariants and contracts preserved? | Yes. Postcondition established: after `task_write` the column equals what `reindex` would derive from the file `task_write` just wrote, asserted in all 4 tests against a real `TaskManager.reindex()` rather than a hand-computed expectation. No lock, ordering or atomicity contract is touched: one local expression between the existing file write and the existing single upsert, and every early return precedes it. The file/DB pair is still not atomic, disclosed as unchanged in base and fix alike. |

Regression: full suite 7 failed / 2937 passed, failure **name set** byte-identical to the
7 pre-existing failures at this base (0 new, 0 disappeared, +4 = the new tests). `ruff`
clean on both changed files.

</details>

oranjeai:
<details>
<summary>Internal second-model review (Gate B) and reviewer adjudication</summary>

Before this PR was opened it went through an independent code review by a second model
(codex, `xhigh` reasoning, cold: it saw the diff and the PR contract, never the author's
evidence), plus a separate reviewer pass that read the changed files and every sibling
task-mutation path from scratch. Both passes ran against commit `fd7f0fc` /
tree `0e0e236b`, and the review scope was checked against the intended change set
(`2 files changed, 183 insertions(+), 1 deletion(-)`) before any finding was adjudicated.

**Outcome: 0 blockers, 0 majors, 4 nits agreed.** One IMPL round, one review round.
Gate spend: $14.45 over 3 runs (2 approach + 1 code).

|---|---|---|---|
| 1 | Test module docstring restates PR rationale rather than harness behaviour | nit | AGREE, noted-not-blocking |
| 2 | `TaskManager.reindex` cited as if it were a live path | nit | AGREE, fixed in the description |
| 3 | The clearing direction's effect on reminders was not disclosed | nit | AGREE, fixed in the description |
| 4 | The PATCH route resolves the same class in the opposite direction | nit | AGREE, follow-up filed |

**1 - test docstring verbosity** (raised by both passes independently). Measured rather
than accepted: prose density is 24.4% of lines (44 of 180), against a median of 13.1%
across the 49 `tests/*.py` files of 150-400 lines. Inside the existing distribution
(the top of it is 39.3%), and every block names a discriminator the tests would be
vacuous without, so this did not justify another round on comments alone. Recorded as a
follow-up; a trim must keep the rival-hook and oracle rationale.

**2 - the indexer is a definition, not a live path.** `TaskManager.reindex` has no caller
anywhere outside tests: excluding the memU bridge and `tests/`, the whole tree yields only
its own definition, its log line, and two migration comments. The description now says so
explicitly. The root-cause argument does not depend on it - it rests on the column having
exactly one SQL writer, no row-only setter (unlike `status` and `tags`, which each have a
dedicated `update_task_*` method), and every one of the seven `upsert_task` suppliers
writing or preserving the file's line. Each was re-verified in this review.

**3 - the direction this change introduces.** The draft stated the escalation consequence
of a resurrected deadline but not of a cleared one. `escalation.py` skips any row whose
deadline is falsy, so clearing the column stops reminders for that task until a
`**Deadline:**` line returns. Both halves are now in the description. Fitting it inside the
description budget was funded by condensing the symptom paragraph, which restated the root
cause; nothing load-bearing was dropped.

**4 - the same class, resolved the other way.** The PATCH route in `routes/tasks.py` is
named in the description as carrying the identical snapshot echo. A separate in-progress
change to that route omits the deadline from its write set when the content omits the line,
which *keeps* the row value where this change *clears* it. Nothing to change here (different
route, and file-authoritative is the defensible direction for a handler that has just
replaced the entire file), and the description's claim is accurate as written. A follow-up
records it so the class is not closed inconsistently.

**Two further reviewer observations were dropped rather than published**, both out of
contract: one duplicate id, and one degenerate parse shape (a `**Deadline:**` line with only
trailing whitespace and nothing after it) where the handler writes NULL and the indexer
writes an empty string. Measured, that shape behaves identically before and after this
change - the key is present, so the replaced fallback is never consulted - so it is
pre-existing and untouched here. It has no effect on the column's only reader, which treats
both as absent.

</details>

oranjeai:
@arsenmuk could you take a look? Small task-store change: `task_write` was re-syncing the DB `deadline` column from a snapshot taken before it replaced the file, so a row could assert a deadline the file it had just written denied. The fix takes the value from the content instead (+3/-1 in `handlers/tasks.py`), plus a regression test.

Requesting via comment because the review-request API returns 404 for this repo (no push access here). If someone else is a better fit for the task store, please redirect.

oranjeai:
### Follow-up on the `deadline` projection: the class is closable, and the direction is measurable

The description of this PR says the PATCH route in `routes/tasks.py` carries the same snapshot
echo and is not closed here. I investigated that residue and want to record two corrections
plus a recommendation, so the class does not get closed inconsistently later.

**1. This PR alone arms the inconsistency - no other change is a precondition.** At
`origin/main` both carriers currently keep the column, so main is self-consistent:

| carrier | `origin/main` form | content omits the line |
|---|---|---|
| `handlers/tasks.py:432` (`task_write`) | `frontmatter.get("deadline", task.get("deadline", ""))` | keeps |
| `routes/tasks.py:152` (PATCH content branch) | `fields.get("deadline") or task.get("deadline")` | keeps |

This PR flips the handler to `frontmatter.get("deadline", "")`, i.e. clear. So after it merges the
two endpoints disagree for byte-identical content, whatever else happens to be in flight.

**2. On `origin/main` the route is not a contract fork - it is one more instance of this PR's own
defect.** `... or task.get("deadline")` reads a pre-write DB snapshot *after* the route has already
written the whole file. That is exactly the staleness this change removes from the handler. (An
in-progress change to that route replaces the line with a deliberate omission, which is a genuine
partial-update contract choice - but that is a property of that branch, not of main.)

**3. File-authoritative looks right for this route specifically.** Four measurements:

- The route's only production caller is a whole-file editor. `GET /api/tasks/{id}` returns the
  entire markdown as `content`; `TaskDetailPage.tsx:157` renders it in a `<textarea>`;
  `handleSave` (`:44`) to `taskStore.saveTaskContent` (`:131-134`) to
  `api.updateTask(id, { content })`. The typed client surface is `{ status?, note?, content? }`
  with no `deadline` field, so deleting the `**Deadline:**` line and saving is a caller's only way
  to express "no deadline" - and keeping silently leaves the row asserting the old value.
- The column has no row-only writer: every SQL statement touching `deadline` is `upsert_task`,
  whose callers all supply it from the file they just wrote (`update_task_status` and
  `update_task_tags` are targeted updates that do not mention it). That is this PR's own accepted
  argument, and it transfers to the route unchanged.
- `escalation.py:65-67` treats an absent and an empty deadline identically, so clearing cannot
  introduce a new escalation behaviour; only keeping can sustain a reminder for a deadline the
  file no longer states.
- The one-line change is `deadline=fields.get("deadline")`.

**4. One correction to an argument I have seen used for this direction.** `TaskManager.reindex` is
the repo's declared file-to-row definition, but it has zero callers outside its own `def` at
`origin/main` (the only other mentions are two migration comments). It is worth citing as declared
intent and as a test oracle, not as live behaviour.

**Why I am not opening a PR for it.** The line is inside the edit scope of an in-progress change
to that route, and a prototype fix based on `origin/main` conflicts with it (and with a second
in-flight change to the same file). More importantly, the surviving form there keeps, so resolving
that conflict by taking either side silently decides the contract. The cheap path is for that
change to project the file's value directly, where the line actually lives. Severity is low
either way: a census of all live task rows found none whose row asserts a deadline its file omits.

No change is requested on this PR - its own direction is the defensible one, and its description
is accurate as written.

pufit (approved):

PR #262, opened 2026-08-04
Flag task_update refusals with is_error so callers can detect them

oranjeai:
## Symptom

`task_update_handler` reports two refusals by **returning** a `ToolResult` rather than raising: an unconfigured status (via `_validate_status`) and a task row that does not exist. Both used `ToolResult.text(...)`, whose `is_error` defaults to `False` (`registry.py:77`), and `ToolRegistry.invoke` does not wrap handlers. So a refusal was indistinguishable from a success on the only machine-readable field a caller has; only string-matching prose could tell them apart. Six call sites discard the result entirely. Both early returns also precede the file write, so the note is lost too.

Reachability is narrow but real: `in_progress` is seeded `is_system=0` and so is deletable, and the missing-task branch needs no config change at all.

## Root cause

`ToolResult` is the model-facing return type. On the agent-loop path a refusal must come back as readable text so the model can self-correct, and `_validate_status` returns a reminder listing the valid statuses. The handler was written for that path only, and `is_error`, the one field carrying "this did not happen" across the boundary, was left at its default.

## Fix

Set `is_error=True` on the three refusal returns in `nerve/agent/tools/handlers/tasks.py`: the unconfigured-status and missing-task returns in `task_update_handler`, and the missing-task return in `task_done_handler`, which `task_update` reaches by delegation when `status == "done"`. Without that third site the flag would be absent for exactly one status. `+3/-3`, one production file.

Raising would have been wrong, since neither refusal is a programmer error. `is_error=True` on a refusal is already the convention here (32 sites across five sibling handler modules). The flag is orthogonal to the text on all three surfaces (`to_dict`, the MCP server's `isError=`, the SDK adapter), so the agent loop keeps reading the reminder verbatim.

**No programmatic caller acts on the flag yet**, so the refused mutation itself is unchanged. The flag does reach the agent loop, where a refusal now records `success=0` in its MCP-usage row (`engine.py:1645`) and renders as an error in the chat UI -- both of which it should. The `PATCH /api/tasks` ordering and the plan-approval boundary are tracked separately and are not addressed here.

## Tests

The existing unconfigured-status test gains an `is_error` assertion and **keeps** its text assertion, which pins the agent-loop contract and would catch a "fix" that flagged by replacing the reminder. Three tests are added: missing task, done-routed missing task, and a success control that must stay unflagged.

Each refusal test fails when its own single token is reverted, and in the `task_done_handler` revert arm the other two still pass, so each test isolates the carrier it names. Full suite compared by sorted FAILED-name sets: identical to `main` (7 pre-existing failures, in two files neither touched here), `+3` passed. `ruff` clean.

### DIFF
--- a/nerve/agent/tools/handlers/tasks.py
+++ b/nerve/agent/tools/handlers/tasks.py
@@ -284,7 +284,7 @@ async def task_update_handler(ctx: ToolContext, args: dict) -> ToolResult:
     if status:
         err = await _validate_status(ctx, status)
         if err:
-            return ToolResult.text(err)
+            return ToolResult.text(err, is_error=True)
 
     # Route done transitions through task_done to ensure file move + FTS sync
     if status == TERMINAL_STATUS:
@@ -293,7 +293,7 @@ async def task_update_handler(ctx: ToolContext, args: dict) -> ToolResult:
     if ctx.db:
         task = await ctx.db.get_task(task_id)
         if not task:
-            return ToolResult.text(f"Task not found: {task_id}")
+            return ToolResult.text(f"Task not found: {task_id}", is_error=True)
 
         new_tags_str = ""
         if raw_tags:
@@ -454,7 +454,7 @@ async def task_done_handler(ctx: ToolContext, args: dict) -> ToolResult:
     if ctx.db:
         task = await ctx.db.get_task(task_id)
         if not task:
-            return ToolResult.text(f"Task not found: {task_id}")
+            return ToolResult.text(f"Task not found: {task_id}", is_error=True)
 
         # Done is a write like any other — it copies the file into done/ and
         # unlinks the source, so a stored ``file_path`` inside the tracked config
--- a/tests/test_task_statuses.py
+++ b/tests/test_task_statuses.py
@@ -133,9 +133,41 @@ class TestStatusToolHandlers:
         await task_create_handler(ctx, {"title": "Some task", "content": "x"})
         task_id = (await db.list_tasks(status="all"))[0]["id"]
         result = await task_update_handler(ctx, {"task_id": task_id, "status": "nope"})
+        # Text is the model-facing reminder and must not move.
         assert "Invalid task status: 'nope'" in _text(result)
+        # is_error is the only field a programmatic caller can test.
+        assert result.is_error is True
         # Status unchanged.
         assert (await db.get_task(task_id))["status"] == "pending"
+
+    async def test_update_missing_task_is_flagged(self, db: Database, tmp_path):
+        ctx = self._ctx(db, tmp_path)
+        result = await task_update_handler(
+            ctx, {"task_id": "no-such-task", "status": "in_progress", "note": "n"},
+        )
+        assert "Task not found: no-such-task" in _text(result)
+        assert result.is_error is True
+
+    async def test_update_missing_task_done_routed_is_flagged(
+        self, db: Database, tmp_path,
+    ):
+        # status == "done" delegates to task_done_handler, a separate return site.
+        ctx = self._ctx(db, tmp_path)
+        result = await task_update_handler(
+            ctx, {"task_id": "no-such-task", "status": "done", "note": "n"},
+        )
+        assert "Task not found: no-such-task" in _text(result)
+        assert result.is_error is True
+
+    async def test_update_success_is_not_flagged(self, db: Database, tmp_path):
+        ctx = self._ctx(db, tmp_path)
+        await task_create_handler(ctx, {"title": "Happy path", "content": "x"})
+        task_id = (await db.list_tasks(status="all"))[0]["id"]
+        result = await task_update_handler(
+            ctx, {"task_id": task_id, "status": "in_progress"},
+        )
+        assert result.is_error is False
+        assert (await db.get_task(task_id))["status"] == "in_progress"
 
     async def test_update_accepts_valid_status(self, db: Database, tmp_path):
         ctx = self._ctx(db, tmp_path)


oranjeai:
<details>
<summary>Internal second-model review (4 independent gate runs, 6 distinct findings, all agreed, 1 fix round)</summary>

An independent model reviewed this change four times: three passes over the approach
before any code was written, and one over the finished diff. A cold reviewer pass over
the resulting code ran twice on top of that. Every finding below was adjudicated with a
citation or a measurement, never prose. Six distinct findings, all agreed; none
disagreed, none left unresolved. Two of the six were raised twice, which under the
fleet's rules removes any option to override them.

|---|---|---|---|
| 1 | The commit message claimed "no caller acts on the flag yet, so no observable behaviour moves". The agent loop does consume `is_error`: a refusal now records `success=0` with its text in the `error` column of its `mcp_tool_usage` row (`engine.py:1645-1646`, inside `if tool_use_id:` and not gated on `not is_error`), and the chat UI renders it as an error. Direction: the claim understated the change, and both movements are the correct reading of a refusal. | ❌ major | AGREE, fixed |
| 2 | `is_error=True` on a refusal was described as the convention at "33 sites across six handler modules". Re-derived at the base commit: 32 sites across five (`workflow_runs` 10, `review_loops` 10, `config_pr` 6, `wakeups` 4, `memory` 2; the other eight handler modules are 0). The 33rd site is `mcp_server/server.py:116`, which is not a handler module. | 💡 nit | AGREE, fixed |
| 3 | The chat-UI citation named `TaskToolBlock.tsx:96/153/165`. `:153` cannot fire for either tool this change touches: it is the else-arm of a ternary guarded by `!isDone && !isUpdate`, and `isUpdate` / `isDone` are true for `task_update` / `task_done`. Reachable sites are `:96` (red icon) and `:165` (the extra error block). Direction: overstated the UI reach by one line. | 💡 nit | AGREE, corrected here |
| 4 | Flagging `task_done_handler` also flags the `task_done` tool's own missing-task refusal, whose three direct callers (`handlers/plans.py:340`, `routes/plans.py:85`, `task_create_handler:238`) likewise discard the result. Correct on that route too, so the widening is in the right direction; the description just did not disclose it. | 💡 nit | AGREE, disclosed |
| 5 | Design pass, raised twice: the first two drafts patched a consumer one layer in rather than the producer. Measuring whether the producer flag alone is behaviourally neutral (it is: across four arms, every observed column except `is_error` is identical) is what narrowed this to `+3/-3` in one file. | ❌ blocker | AGREE, fixed |
| 6 | Design pass, raised twice: disclosing a residual is not owning it. The `PATCH /api/tasks` write ordering and the plan-approval compensation boundary each became their own tracked work with a measured readiness gate, rather than riding along here. | ❌ major | AGREE, fixed |

**Independent-model pass on the final code: 0 findings.** It read both changed files in
full, every `task_update` / `task_done` caller, the `ToolResult` adapters and consumers,
the refusal siblings and the tests.

**Verified by the reviewer, not relayed.** All 30 `ToolResult.text` returns in the
changed file were enumerated with their owning function, confirming the three refusal
carriers reachable from `task_update` are complete and that the other returns in those
functions are success returns. Test liveness was traced rather than assumed: `:296` and
`:457` are byte-identical lines, so the two missing-task tests are discriminated only by
owning function, and the success control blocks a flag-everything fix. The kept text
assertion is the agent-loop contract in test form. No other test can break: the lockdown
paths raise rather than return, nothing asserts `is_error is False` on a task refusal,
and the one web-side matcher of "Task not found" is a bare anchored regex from a
different producer that cannot match this colon-and-id text.

**Residuals, stated plainly.** Findings 3 and 4 were corrected on this comment but not in
the commit message, which still cites `:153` and still frames the third carrier as
delegation-only. Both understate nothing that affects the merge decision, and a full
implement-plus-review round for two text qualifiers on a test-adjacent change is
manufactured work; a follow-up is tracked. Nothing else was left unresolved.

</details>

oranjeai:
<details>

| # | Check | Answer |
|---|---|---|
| a | **Deterministic repro?** | Yes. A 4-arm probe against an isolated tmpdir (`Database(path)+connect()`, `ToolContext` built as in `tests/conftest.py`) reproduces on demand, no CI and no server: `invalid-status`, `missing-task`, `done-routed` and a valid control all return `is_error=False`, so a refusal is indistinguishable from a success. Not a "~1%" flake. |
| b | **Root cause explained?** | Yes. `ToolResult` is the model-facing return type, so `task_update_handler` reports both refusals by returning readable text instead of raising; `ToolResult.text(...)` leaves `is_error` at its `False` default (`registry.py:77`) and `ToolRegistry.invoke` is a bare `await spec.handler(ctx, args)` with no wrapping, so the field that would carry "this did not happen" is never set. |
| c | **Fix matches the root cause?** | Yes. It sets that one field, at the producer, on the refusal branches only. No widened bound, no defensive check masking an upstream bug, no re-reading of state to infer what the callee already computed. |
| d | **Test intent preserved / new tests added?** | Yes. The existing unconfigured-status test keeps its text assertion, which pins the agent-loop contract and would catch a "fix" that flagged by *replacing* the reminder, and gains an `is_error` assertion. Three tests added: missing task, done-routed missing task, and a success control that must stay unflagged. |
| e | **Demonstrated in both directions?** | Yes. Base tree vs an export carrying only the three tokens (mutation asserted present: 3 hits; absent at base: 0): the three refusal arms flip `False -> True`, the success control stays `False`, and every other observed column (raised, resulting status, note-present) is identical. Plus a 3-mutant matrix, each reverting one token, with an unmutated control green at both ends and the tree verified restored afterwards. |
| f | **General across code paths?** | Yes. All 30 `ToolResult.text` returns in the file were enumerated with their owning function via AST. The three refusal carriers reachable from `task_update` are fixed, including `task_done_handler`, which `task_update` reaches by delegation when `status == "done"`; without it the flag would be absent for exactly one status. The remaining `task_update_handler` / `task_done_handler` returns are *success* returns ("updated." / "marked as done.") and are correctly unflagged; the `LockdownError` paths already raise. Flagging `task_done_handler` also flags the `task_done` tool's own missing-task refusal, whose three direct callers (`handlers/plans.py:340`, `routes/plans.py:85`, `task_create_handler:238`) likewise discard the result; a missing task is a refusal on that route too, so the widening is in the right direction. The `task_create` / `task_read` / `task_write` refusals and the same pattern in `handlers/plans.py` belong to different tools and are named as out of scope rather than overlooked. |
| g | **Generalizes across inputs?** | Yes. The flag is set unconditionally on each refusal branch, so no input can reach the branch and skip it. The refusal text is an f-string over `task_id` / `status` and `is_error` is orthogonal to it. No type-wrapper or boundary-value matrix applies: there is no data path and no parsing here. |
| h | **Backward compatible?** | Yes. No setting, no serialization format, no migration. `is_error` already crosses both wire surfaces (`to_dict`, `CallToolResult(isError=...)`). Every consumer was grepped (agent engine, MCP server and audit, the Claude backend, the web stores) and each treats it as a presentation or bookkeeping flag; none treats it as fatal. No consumer string-matches the refusal text programmatically: the one web-side matcher is a bare anchored `/^\s*Task not found\s*$/i`, measured not to match this colon-and-id text. |
| i | **Invariants and contracts preserved?** | Yes. No control flow moved: same branch, same text, same early-return position, so the pre-existing behaviour that the note is dropped on refusal is unchanged (fixing that is a separate concern). The agent-loop contract holds because `content` is untouched. No lock, atomicity or durability path is involved, since nothing is persisted on either refusal branch. |

**Scope.** No *programmatic* caller acts on the flag yet, so the refused mutation itself is unchanged. The flag does reach the agent loop: a refusal now records `success=0` (and its text in `error`) in the `mcp_tool_usage` row (`engine.py:1645-1646`, reached because the tool is exposed as `mcp__nerve__task_update`), and the chat UI renders it as an error (`streamingHandlers.ts:343`, `TaskToolBlock.tsx:96/165`). Both are the correct reading of a refusal, but they are behaviour changes and the earlier blanket "nothing moves" wording was wrong. The `PATCH /api/tasks` write ordering and the plan-approval compensation boundary are tracked separately and are deliberately not addressed here.

**Regression check.** Full suite compared by sorted FAILED-test-name sets rather than counts: identical to `main` (7 pre-existing failures in `test_memu_bridge.py` and `test_telegram_sessions.py`, neither file touched by this change), `+3` passed, 0 collection errors. `ruff` clean on both changed files; all added lines ASCII-only.

</details>

oranjeai:
### Correction to the commit message and to my earlier comment on this PR

Following up on the two disclosed nits in my adjudication comment above. I
re-derived both against the tree and one of my own corrections was itself
incomplete, so this supersedes the relevant rows. No code change: the diff
stays `+3/-3` in `nerve/agent/tools/handlers/tasks.py` and the tree is
unchanged.

**1. The chat-UI citation is `TaskToolBlock.tsx:96/159/165`.**

The commit message says `:96/153/165` and my earlier comment corrected that to
`:96/165`. Both are wrong. `:153` is indeed unreachable for the two tools this
change touches -- it is the else-arm of a ternary guarded by
`!isCreate && !isDone && !isUpdate`, and `isUpdate` / `isDone`
(`TaskToolBlock.tsx:69-70`) are true for `task_update` / `task_done`, while the
sibling arm needs `taskList.length > 0` and `taskList` is only populated when
`isList` (`:86`). But dropping it left out a third reachable site.

`:159` is the success banner, guarded by
`(isCreate || isDone) && resultText && !block.isError && !taskList.length`.
Before this change, a `task_done` refusal (`Task not found: ...`) satisfied it,
so the UI rendered a green icon at `:96`
(`block.isError ? red : isDone ? 'text-hue-green' : ...`) together with the
literal string `Task completed` at `:161` -- for a task that does not exist and
was not completed. Setting `is_error=True` suppresses `:159` and turns `:96`
red.

Modelled for every tool that reaches these three refusal sites, with
`resultText` non-empty (a refusal always carries text):

| tool | `is_error` | `:96` icon | `:153` | `:159` banner | `:165` |
|---|---|---|---|---|---|
| `task_update` | False | muted | no | no | no |
| `task_update` | True | red | no | no | yes |
| `task_done` | False | **green** | no | **yes** | no |
| `task_done` | True | red | no | no | yes |

So the change affects `:96` and `:165` for `task_update`, and `:96`, `:159` and
`:165` for `task_done`. The `web/` tree is untouched by this PR, so these line
numbers hold identically at the base commit `94406ea`.

**2. `task_done_handler`'s missing-task refusal also serves the `task_done`
tool directly.**

The message frames that third site purely as the one `task_update` reaches by
delegation when `status == "done"` (`tasks.py:291`). That is true but
incomplete: the same function is the handler of `TASK_DONE_SPEC`
(`tasks.py:599-604`), exported in `TASK_SPECS` (`:635`), so the refusal is also
the `task_done` tool's own. Its three other callers all discard the result --
`handlers/plans.py:340` (plan decline), `gateway/routes/plans.py:85-89` (plan
decline over HTTP, via `invoke("task_done", ...)`), and `tasks.py:238`
(`task_create` in the terminal status). A missing task is a refusal on those
routes too, so the flag is right there as well; this only widens the
description.

Both corrections **understate** what the change does rather than overstate it,
so neither affects the merge decision. I am leaving `9aa9141` as it is rather
than amending or adding an empty commit: this repo's convention is to add new
commits instead of rewriting, and a zero-tree commit would not appear in
`git log -- <path>` for either affected file, which is exactly where a future
reader would look. If a substantive round happens on this PR, I will fold both
corrections into that commit's message.

pufit (approved):

PR #264, opened 2026-08-04
Report a dead runtime instead of returning an empty turn

oranjeai:
## Symptom

A claude turn whose runtime exits without producing a result is recorded as a normal,
successful turn: no error, no retry, an empty assistant message. The engine's next-turn
health check then replaces the dead client, so the damage is one silently-empty turn.

## Root cause

`base.py` specifies `receive_turn` as "yield events until (and including)
`TurnCompleted`", which `events.py` calls "always emitted exactly once per turn". On
`StopAsyncIteration` from the SDK response iterator, `receive_turn` did a bare `return`
yielding no terminal event, so the engine's receive loop fell out of its `async for` and
reached `break  # success`.

That EOF is not the crash path it looks like. An SDK exception, including `ProcessError`
for a non-zero exit, arrives as an error message and *raises*; the transport generator's
only non-exception exit is falling off the end after `await process.wait()` returned rc
0. So the branch means one thing: the runtime exited without a result, already dead.

## Fix

Raise `TransportDiedError` there instead of returning. The contract is then satisfied
through the protocol's own error channel, and the engine's existing dead-client path
handles it: with no content yet it discards the client and retries fresh, and with content
streamed it re-raises to the error handler. Both replace the client, so a spent stream
cannot be reused. The codex backend does this already for the same condition (`d42635a`).

Two consequences are deliberate. A previously-silent turn now surfaces an error. And with
partial content the live view replaces the streamed text with the error, while the stored row
keeps that text in `blocks` and the error in `content`, so a reload shows the text, not the
error. That split is how every mid-turn transport death already behaves. Synthesizing a failed
`TurnCompleted` would avoid the live replacement, but reports a dead transport as completed
and leaves the client in the pool. Happy to switch.

## Tests

Eight new tests: the EOF raise with and without prior content, a completed turn still
yielding exactly one `TurnCompleted`, the idle timeout asserted *not* a `TransportDiedError`,
a real stream error propagating unrelabelled, iterator cleanup on both paths, and both
sides of the engine's retry gate through the real receive loop.

Two existing tests are edited; both **passed on main**, so they are worth a look.
`test_receive_turn_handles_empty_stream` asserted `seen == []`: it pinned this defect, so it
must fail under any repair; it becomes a `pytest.raises`, keeping its cleanup assertion.
`test_receive_turn_yields_events_normally` ended its stub stream without a `ResultMessage`,
so it now hits the raise; it gains one (reverting it reproduces the failure).

Full suite: identical FAILED-test sets before and after (7 pre-existing, unrelated),
2933 to 2941 passed. `ruff`: no new findings. A 7-mutant matrix covers the branch, the
exception *type*, the sibling error paths, the cleanup and the split above.

### DIFF
--- a/nerve/agent/backends/claude.py
+++ b/nerve/agent/backends/claude.py
@@ -1025,7 +1025,13 @@ class ClaudeClient(AgentClient):
                     else:
                         message = await response_iter.__anext__()
                 except StopAsyncIteration:
-                    return
+                    # The stream ended with no result and no error, which
+                    # means the runtime already exited (a non-zero exit
+                    # raises instead). base.py requires a terminal event,
+                    # so report the death rather than a silent success.
+                    raise TransportDiedError(
+                        "claude runtime ended the turn without a result",
+                    ) from None
                 except asyncio.TimeoutError:
                     logger.warning(
                         "CLI idle timeout (%ds) for session %s — no SDK "
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -2,14 +2,16 @@
 helpers (no SDK subprocess)."""
 
 import asyncio
+import json
 import os
 from types import SimpleNamespace
 from unittest.mock import AsyncMock, patch
 
 import pytest
-from claude_agent_sdk import AssistantMessage, TextBlock
+from claude_agent_sdk import AssistantMessage, ResultMessage, TextBlock
 
-from nerve.agent.backends.base import SessionSpec
+from nerve.agent.backends import events as ev
+from nerve.agent.backends.base import SessionSpec, TransportDiedError
 from nerve.agent.backends.claude import ClaudeBackend, ClaudeClient, translate_message
 from nerve.agent.engine import AgentEngine, _TurnState, _model_family
 from nerve.config import AgentConfig, NerveConfig
@@ -196,10 +198,24 @@ def _translated(messages: list) -> list:
     return [event for m in messages for event in translate_message(m)]
 
 
+def _result_msg(session_id: str = "sdk-1") -> ResultMessage:
+    """A terminal ResultMessage: translates to one TurnCompleted."""
+    return ResultMessage(
+        subtype="success", duration_ms=1, duration_api_ms=1,
+        is_error=False, num_turns=1, session_id=session_id,
+        total_cost_usd=0.5, usage={"input_tokens": 1},
+    )
+
+
 @pytest.mark.asyncio
 async def test_receive_turn_yields_events_normally():
-    """Fast SDK stream completes without timing out."""
-    messages = [_text_msg("a"), _text_msg("b"), _text_msg("c")]
+    """Fast SDK stream completes without timing out.
+
+    The stream ends with a ResultMessage: that terminal event is what
+    makes the turn complete, so ``receive_turn`` returns rather than
+    reporting the exhausted stream as a transport death.
+    """
+    messages = [_text_msg("a"), _text_msg("b"), _text_msg("c"), _result_msg()]
     sdk = _StubClient(messages)
     client = _turn_client(sdk, "sess-1", idle_timeout=5.0)
     seen = []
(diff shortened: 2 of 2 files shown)


oranjeai:
<details>
<summary>Internal second-model review (5 findings across 5 gate rounds, all agreed and fixed; 0 open)</summary>

Before this PR was opened it went through five independent review rounds: three
against the plan (before any code existed) and two against the code, each by a
separate model that did not see the reasoning behind the change, plus my own cold
reviews. Every finding is listed with its verdict. The final round found nothing.

| # | Finding | Severity | Verdict | Resolution |
|---|---|---|---|---|
| 1 | `design-1` - the first approach synthesized a failed `TurnCompleted` instead of raising, on a reachability claim about a dropped `end` sentinel that does not hold | blocker | AGREE | The approach was replaced before any code was written. Reaching that sentinel requires `process.wait()` to have returned, so the process is already dead and a retry cannot re-prompt a live CLI. |
| 2 | `band-aid-1` - a spent stream must be visible to `is_alive()` | major | AGREE (superseded) | Adopted, then removed by finding 3 once the design raised instead of returning. Recorded rather than dropped so a later round does not restore it. |
| 3 | `minimality-1` - the `is_alive()` guard added for finding 2 is unreachable | major | AGREE | Proven unreachable: `receive_turn` has one production caller and both of its failure outcomes already remove and disconnect the client. The guard was deleted, leaving a change of seven production lines. |
| 4 | `partial-eof-oracles` - the partial-content engine test did not prove the promised cleanup or the persisted split | major | AGREE | Confirmed with a mutant: leaking the dead client left both engine tests green. Four assertions were added (client disconnected, client unpooled, persisted `content` carries the error, persisted `blocks` keeps the partial text). Two new mutants now fail against it and pass on shipped code, and both survive against the previous test file, which attributes the coverage to the new assertions rather than to pre-existing tests. |
| 5 | `partial-text-claim-inverted` - the description stated the partial-text trade-off backwards | major | AGREE | The streamed text is replaced in the live view and in the return value, but the stored row keeps it in `blocks` while `content` carries the error, and reload renders assistant messages from `blocks`. So a reload shows the text, not the error. The description, the commit message and an inline comment were all corrected, and a test now pins both halves. |

Two smaller notes, not blocking:

- The symptom wording originally said a later turn stays empty "until the health
  check happens to replace it". That check runs on every turn and reads the
  subprocess return code, which is necessarily set here, so the blast radius is
  one silently-empty turn. Corrected.
- `test_engine_fails_turn_when_eof_follows_content` still opens its docstring
  with "Two consequences are asserted deliberately" while the body now carries
  six assertions. The two it names are the deliberate trade-offs and nothing it
  says is false; only the count is stale. Left as is rather than spending a
  round on a comment.

Validation the reviews asked for: both directions in three isolated arms (fix
plus new tests, base source plus new tests, base source plus base tests), a
seven-mutant matrix that is fully killed with an unmutated control green at both
ends of the run, identical whole-suite failure sets before and after, and no new
`ruff` findings.

Review spend: $22.36 across 5 gate runs (3 on the plan, 2 on the code).

</details>

oranjeai:
<details>

| # | Question | Answer |
|---|---|---|
| a | Deterministic repro? | Yes. `pytest tests/test_engine.py -k "test_receive_turn or test_engine_retries_fresh or test_engine_fails_turn"`. The stub stream exhausts immediately, so the EOF branch is entered on every run: not probabilistic. |
| b | Root cause explained? | Yes. `receive_turn` bare-returned on `StopAsyncIteration`, yielding no terminal event, so the engine's receive loop reached `break  # success`. That EOF can only follow `await process.wait()` returning rc 0 (an SDK exception, incl. `ProcessError` on non-zero exit, raises instead), so it means the runtime exited without a result. |
| c | Fix matches the root cause? | Yes. The condition is a dead transport and `TransportDiedError` is the type for that; the receiving machinery already exists. ~7 lines in one `except` branch, no new state, no new import, no defensive check. |
| d | Test intent preserved / tests added? | Yes. 8 new tests. Two existing tests are edited; both passed on main, both are forced by the fix, and the rewritten one keeps its cleanup assertion. Called out in the description. The partial-content engine test also now asserts the dead client is disconnected and unpooled and inspects the persisted row, which is what makes "a spent stream cannot be reused" falsifiable rather than stated. |
| e | Demonstrated in BOTH directions? | Yes, three stash-isolated arms: fix + new tests = 12 passed; base source + new tests = 3 failed (`DID NOT RAISE TransportDiedError`); base source + base tests = 4 passed, including the old `seen == []` assertion. |
| f | General across CODE paths? | Yes. `CodexClient.receive_turn` needs no change (no EOF path; already raises on observed death) and this makes claude match it, per `d42635a`. The idle-stream readers on the same per-client stream already treat `end`/`error` as over. Fixed at the backend seam, not guarded at the engine. |
| g | Generalizes across INPUTS? | The branch takes no user input; the relevant matrix is stream shapes, all covered: empty, text-then-EOF, text-then-result, hang (timeout), iterator raising its own error, plus both sides of the engine's `got_content` gate. |
| h | Backward compatible? | No setting, serialization or schema surface. It is a behaviour change for every session, which the description states plainly, along with the partial-text split (the live view and the return value carry the error, while the stored row keeps that text in `blocks`, so a reload shows the text and not the error) and the alternative that would avoid the live replacement. |
| i | Invariants and contracts preserved? | Yes. The terminal-event contract is met via the error channel; exactly-once cannot double-fire (the raise yields nothing); cleanup runs exactly once on the raise path; the idle timeout stays a distinct signal (asserted *not* a `TransportDiedError`) and real errors propagate unrelabelled. Resume is preserved: the raise precedes client removal, and a test asserts the persisted `sdk_session_id` after the failure. The engine's error path is asserted to remove and disconnect the dead client and to leave the partial text in the stored `blocks`, so both halves of the durable split are pinned. |

Also measured: full suite FAILED-test sets are byte-identical before and after (7 pre-existing, unrelated); passing 2933 -> 2941. `ruff --output-format=concise`: 79 pre-existing findings on both arms, 0 new, 0 in either touched file. A 7-mutant matrix (revert the branch; raise a bare `Exception`; widen the `except`; swallow other errors; delete the `finally`; leak the dead client; clear the persisted partial text) is 7/7 killed with the unmutated control green at both ends of the run. The last two are killed only by the engine assertions added in review round 1: against the previous test file both survive, which is what attributes the coverage to those assertions rather than to pre-existing tests.

One note worth recording: the obvious cleanup assertion for this path is vacuous. `_StubClient.aclose_calls` counts its own generator's `finally`, which runs on exhaustion and on cancellation whether or not `receive_turn` closes the iterator, so it stays 1 even with the `finally` deleted (the pre-existing timeout tests share that blind spot). The mutant that deletes the cleanup survived until two tests were added that count a real `aclose()` call.
</details>

pufit (approved):

PR #279, opened 2026-08-05
[7/7] Reorder board columns by dragging

alex-clickhouse:
> **Stacked on #277** → #276 → #275 → #274 → #273 → #272 → `main`.

From live testing feedback. Column order was fixed at whatever `sort_order` the statuses happened to have, with no way to change it from the UI — and lane order is the part of a board people arrange to match how they actually work.

## Two details make the drag work

Columns are sortable **in the same `DndContext` as the cards**, branching on what was picked up rather than on what it landed on.

1. **Drag listeners live on the header only**, not the whole column. The body has to stay a drop target for cards; listeners on the wrapper would make every card drag also pick up its column.
2. **Column ids are namespaced** (`col:<status>`) so they can't be confused with a task id inside the shared context — and a drop resolves through *any* kind of target: another column, a lane body, or a card. Collision detection returns the nearest droppable of any kind, so a column drag routinely finishes with the cursor over a card. Treating that as "no target" would make the gesture feel broken for no reason.

## The API takes a whole sequence, not an index

`POST /api/task-statuses/reorder {names: [...]}`.

A drag shifts several statuses at once, so a full-sequence write is idempotent and avoids the half-applied states a per-row `PATCH` storm would leave behind if one of them failed. Statuses omitted from the request keep their relative order *after* the named ones, so an incomplete call can't drop a column off the board.

Applied optimistically in the store so the dragged column doesn't snap back mid-request; the server's echo is authoritative and a failure reloads rather than guessing.

## Testing

- [x] `pytest tests/ -q` — **3113 passed**
- [x] `npm test` — **84 passed** (11 for the column geometry, 3 for the column announcements)
- [x] `tsc -b` + `npm run build` clean · `eslint` 146

The reorder route test includes a declaration-order guard — `/reorder` has to stay above `/task-statuses/{name}` or it gets captured as a status name.

**Not verified: the drag gesture itself in a browser.** Geometry and the reorder reducer are tested as pure functions, but dnd-kit's sensor behaviour with two nested `SortableContext`s (horizontal columns, vertical cards) isn't. Needs a `nerve restart` on the instance to deploy the endpoint before I can check it.

### DIFF
--- a/nerve/db/task_statuses.py
+++ b/nerve/db/task_statuses.py
@@ -68,6 +68,34 @@ class TaskStatusStore:
         ) as cursor:
             return [dict(row) async for row in cursor]
 
+    async def reorder_task_statuses(self, names: list[str]) -> list[dict]:
+        """Rewrite ``sort_order`` to match the given sequence of names.
+
+        Takes the whole desired order rather than one status's new index:
+        board columns are reordered by dragging, which shifts several
+        statuses at once, and sending the full sequence makes the write
+        idempotent and free of the intermediate states a per-row PATCH
+        storm would leave behind if one of them failed.
+
+        Names not present in the list keep their relative order after the
+        ones that are, so an unknown or omitted status can't be silently
+        dropped from the board.
+        """
+        async with self._atomic():
+            async with self.db.execute(
+                "SELECT name FROM task_statuses ORDER BY sort_order ASC, name ASC"
+            ) as cursor:
+                existing = [row[0] async for row in cursor]
+
+            wanted = [n for n in names if n in existing]
+            ordered = wanted + [n for n in existing if n not in wanted]
+            await self.db.executemany(
+                "UPDATE task_statuses SET sort_order = ? WHERE name = ?",
+                [(i, name) for i, name in enumerate(ordered)],
+            )
+
+        return await self.list_task_statuses()
+
     async def get_task_status_def(self, name: str) -> dict | None:
         async with self.db.execute(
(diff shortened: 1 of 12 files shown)


alex-clickhouse:
**Addressed the suppressed review comment** (the one folded into the review body about drag announcements) in `bac0df9`.

It was right, and it was the same defect Copilot raised on #275 — the announcements block was byte-identical at both commits, so #279 inherited it and then made it worse by putting a non-task into `active`.

Fixed across the two PRs rather than twice in one place:

- **#275** replaced the inline ternaries with a pure `boardAnnouncements(labelFor)` in `announcements.ts`, handling `task` and `lane` — the two types that exist at that commit. Lanes now resolve to the configured *label*, and the raw `lane:` prefix no longer leaks (`onDragEnd` had no `replace` at all, so it announced `Task dropped on lane:pending.`).
- **This PR** adds the `column` branch, which is the right place for it since this is the PR that introduces column dragging. Without it a column drag announced `Picked up task col:pending.`

It reads from `active.data.current.type` as suggested; resolving through `statusFromDropTarget` wasn't needed because the column droppable already carries `{ type: 'column', status }` in its payload, same as lanes do.

Three new cases in `announcements.test.ts` cover column-picked-up, column-over-column, and column-dropped-on-a-card — that last one because collision detection returns the nearest droppable of any kind, so a column drag routinely finishes with the cursor over a card.

`npm test` 84 passed · `pytest` 3113 passed · `tsc -b` + build clean · eslint 146 (down from 148 on this branch before the review pass).

pufit (approved):

PR #304, opened 2026-08-10
Telegram /sessions: paginate the switcher, pin starred, exclude cron

serxa:
## Problem

On Telegram, `/sessions` surfaced only a handful of recent sessions, and starred (kept-alive) sessions were usually *not* among them:

- The switcher fetched the 30 most-recently-updated sessions across **all** sources, then filtered to interactive (telegram/web) ones in Python. Cron/automation sessions update constantly, so they dominated that 30-row window — in practice only a few interactive sessions survived the filter, well under the button cap (a user reported seeing 4).
- Ordering was pure recency, and starring only added a ⭐ to a label *if* the session already appeared. Since starring exists to keep **idle** sessions alive, those sessions have an old `updated_at` and sank below the cutoff — so the sessions a user most wants to return to were the least reachable.
- There was no way to reach anything past the first page.

## Change

- New `SessionStore.list_interactive_sessions(limit, offset, sources, current_id)` that does the work in SQL: filters to interactive sources and non-empty sessions (so cron/automation and 0-message sessions never consume the window), and orders **current → starred → most-recent**. Exposed through `SessionManager` and the channel `Router`.
- `_sessions_view_for` now fetches one page (`page_size + 1`, to detect a further page) instead of over-fetching and counting messages per session in Python.
- `build_sessions_view` renders `⬅️ Prev` / `➡️ More` buttons (new `sess:page:<offset>` callback) so **every** switchable session is reachable. Pinning the current session first keeps it on page 1 regardless of how many starred sessions precede it; starred sessions are pinned ahead of merely-recent ones so an idle kept-alive session is never buried.
- Page size 8 → 10.
- Dropped the per-row ⭐/☆ toggle: Telegram sizes a row's buttons evenly with no API for a narrow second column, so the toggle took half of every session row. The ⭐ stays as a read-only "kept alive" marker in the switch label; star/unstar via `/star` · `/unstar`. Each row is now a single full-width tap-to-switch button.

## Tests

- `tests/test_list_interactive_sessions.py` (new) — current→starred→recent ordering; exclusion of cron / empty / archived; pagination via limit/offset; next-page probe.
- `tests/test_telegram_sessions.py` — updated the switcher tests for the paged, single-button-per-row view; added first/middle/last-page nav assertions.
- Full suite: `3196 passed`, 1 pre-existing unrelated failure (`test_xmemory_bridge` — an `xmemory` package version mismatch, independent of this change).

### DIFF
--- a/nerve/agent/sessions.py
+++ b/nerve/agent/sessions.py
@@ -609,6 +609,21 @@ class SessionManager:
             limit=limit, include_archived=include_archived,
         )
 
+    async def list_interactive_sessions(
+        self,
+        limit: int,
+        offset: int = 0,
+        sources: tuple[str, ...] = ("telegram", "web"),
+        current_id: str | None = None,
+    ) -> list[dict]:
+        """Paginated switchable sessions (current → starred → recent) for the
+        channel /sessions keyboards. See
+        :meth:`nerve.db.sessions.SessionStore.list_interactive_sessions`.
+        """
+        return await self.db.list_interactive_sessions(
+            limit=limit, offset=offset, sources=sources, current_id=current_id,
+        )
+
     async def set_starred(self, session_id: str, starred: bool) -> bool:
         """Star or unstar a session.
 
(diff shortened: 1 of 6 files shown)


pufit:
Amazing feature! Exactly as I imagined it should be implemented

pufit (approved):

PR #318, opened 2026-08-19
Add a scheduled fresh-resolve job to catch dependency drift

alex-clickhouse:
Stacked on #319 — **merge second**. The diff shows two commits; read `d23d438`
alone (a single new workflow file). Based on #319 rather than plain `main` because
this job installs *unpinned*: on a `main` that still lacks the mcp bound it would go
red, correctly, since fresh installs are genuinely broken there.

## The gap

Nerve has no Python lockfile — `pyproject.toml` carries lower bounds only — so what
`uv pip install -e .` resolves changes silently as upstream publishes releases.
`ci.yml` structurally cannot see that drift: it keys its uv cache on
`pyproject.toml`, so a restored cache carries stale index metadata and an unchanged
pyproject means an unchanged resolution, indefinitely.

Not hypothetical. mcp 2.0.0 published 2026-07-28 and broke every fresh install at
startup. CI stayed green for three weeks —
[run 32061993270](https://github.com/ClickHouse/nerve/actions/runs/32061993270)
(main, 2026-08-17) installed `mcp==1.29.0`. It surfaced only when a user filed #316,
and the delayed cost landed on whoever next touched `pyproject.toml`.

The uncomfortable framing: **a green CI run has never been evidence that a fresh
install of Nerve works.** With ~15 lower-bound-only direct dependencies, every one of
them can do what mcp did.

## What this adds

A separate daily workflow that resolves with the cache **off** and index metadata
refreshed, then runs the suite. It answers exactly one question: *does a brand-new
install of Nerve still work today?*

- **Daily at 06:00 UTC** + `workflow_dispatch`. Detects drift within a day for ~2 min of runner time.
- **Publishes the full resolution** (`uv pip freeze`) to the run summary, so drift is reviewable on green runs and diffable between runs when something breaks.
- **On failure, says what it means**: an upstream release broke a fresh install, the tree didn't change, and `ci.yml` can legitimately still be green.

## Deliberate scoping

- **Separate file, not a job in `ci.yml`** — different triggers, and PR runs should stay cached and fast. A `schedule` job inside `ci.yml` would need `if: github.event_name == 'schedule'` guards on every job.
- **Frontend excluded** — `web/package-lock.json` + `npm ci` already pin it reproducibly. The gap is Python-only.
- **`pull_request` trigger scoped to this file's own path** — `schedule` and `workflow_dispatch` only fire for the copy on the default branch, so without it a change to this workflow couldn't be exercised until after merging. It also means future edits here get validated, rather than this being a one-off testing hack.

## Relationship to #320

#320 pins `ci.yml` through a lockfile. That's the other half of the same idea, and the
split is the point: **`ci.yml` pinned** so a PR is never broken by an unrelated
upstream release, **this job unpinned** so drift is still caught somewhere that isn't
someone's PR. Either one alone leaves a gap.

## Testing

- [x] YAML parses; triggers, cron expression, `enable-cache: false`, and step list all verified programmatically
- [x] Ran green on this PR via the self-scoped `pull_request` trigger — and verified from the log that it did what it claims rather than merely passing: `enable-cache: false`, no cache restore attempted, `--refresh` used, `Resolved 120 packages`, `+ mcp==1.29.0` (on the earlier pre-port base)
- [x] The underlying commands are the ones used to diagnose #316: run against an unfixed `pyproject.toml` they resolve `mcp==2.0.0` and the suite dies with 8 collection errors — exactly the signal this job exists to raise

## Not covered

This detects drift; it doesn't prevent it — that's #320.

### DIFF
--- a/.github/workflows/fresh-deps.yml
+++ b/.github/workflows/fresh-deps.yml
@@ -1,0 +1,110 @@
+name: Fresh dependency resolve
+
+# Answers one question the PR/push CI structurally cannot: "does a brand-new
+# install of Nerve still work today?"
+#
+# Nerve has no Python lockfile — pyproject.toml carries lower bounds only — so
+# what `uv pip install -e .` actually resolves changes silently as upstream
+# publishes new releases. ci.yml can't see that drift: it keys its uv cache on
+# pyproject.toml, and a restored cache carries stale index metadata, so an
+# unchanged pyproject means an unchanged resolution indefinitely.
+#
+# That blind spot is not hypothetical. mcp 2.0.0 (published 2026-07-28) removed
+# an API that nerve/mcp_server/http.py imports at module scope, breaking every
+# fresh install at startup — while CI stayed green for three weeks still
+# resolving mcp 1.29.0 from cache. It surfaced only when a user filed #316.
+#
+# So: resolve with the cache off and index metadata refreshed, then run the
+# suite. A failure here means "an upstream release broke a fresh install",
+# NOT "someone's PR broke the build" — the tree is unchanged, only the wider
+# ecosystem moved.
+#
+# The frontend is deliberately out of scope: web/package-lock.json plus `npm ci`
+# already pin it reproducibly. The gap being closed is Python-only.
+
+on:
+  schedule:
+    # 06:00 UTC daily. Detects drift within a day of an upstream release,
+    # for ~2 minutes of runner time.
+    - cron: "0 6 * * *"
+  workflow_dispatch:
+  # Also run when this workflow file itself changes. `schedule` and
+  # `workflow_dispatch` only fire for the copy on the default branch, so
+  # without this a change here couldn't be exercised until after it merged.
+  pull_request:
+    paths:
+      - ".github/workflows/fresh-deps.yml"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: fresh-deps-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  fresh-resolve:
+    name: Fresh resolve + tests (Python ${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        # Mirrors ci.yml. memu-py==1.4.0 requires >=3.13, so 3.13 is the real
+        # floor despite pyproject's requires-python being >=3.12.
+        python-version: ["3.13"]
+    steps:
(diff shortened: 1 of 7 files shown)


alex-clickhouse:
Verified the job does what it claims rather than just going green —
[run 32256072492](https://github.com/ClickHouse/nerve/actions/runs/32256072492):

```
Set up uv    enable-cache: false          ← no cache restore attempted at all
Install      Run uv pip install -e ".[test]" --refresh
Install      Resolved 120 packages in 1.41s
Install       + mcp==1.29.0
Fresh resolve + tests (Python 3.13)  pass  1m43s
```

So it resolved with the cache off and index metadata refreshed, and the suite
passed against that genuinely fresh resolution.

**One thing a reviewer will notice:** only the fresh-resolve check appears on this
PR. `ci.yml` triggers on `pull_request: branches: [main]`, and this PR targets
`alex-clickhouse/pin-mcp-below-2` while stacked. The normal backend/frontend checks
will run automatically once #317 merges and GitHub retargets this to `main`.

**Worth being explicit about what this run does and doesn't prove.** It proves the
job executes correctly and is cache-free. It does *not* prove it catches drift,
because its base already contains #317's bound. The evidence for the catching half
is that these exact steps against the unfixed `pyproject.toml` resolve `mcp==2.0.0`
and the suite dies with 8 collection errors — which is how #316 was diagnosed in the
first place. Put differently: run this job against today's `main` and it goes red,
correctly, because `main` really is broken for fresh installs.

alex-clickhouse:
Closing as a separate PR — folded into #320, which already carries this commit
(`d23d438`) in its history. Nothing is lost: `fresh-deps.yml` still ships, with its
daily schedule intact.

Splitting these stopped making sense once #320 started *editing* this workflow. The
lockfile is what creates the bounds-blindness that the deps-PR trigger fills, so the
two changes only make sense read together — reviewing them as separate PRs meant
reviewing half an argument twice.

**One thing I kept on purpose, in case it's what you meant to drop:** the daily
`schedule` trigger. The PR-scoped trigger only fires when someone touches
`pyproject.toml` or `requirements.lock`, so on its own it would miss the exact
scenario that started all of this — mcp 2.0.0 broke fresh installs on 2026-07-28 and
went unnoticed for three weeks precisely *because* no PR touched dependencies in that
window. The two triggers cover genuinely different failure modes:

- **daily** — drift that happens while nobody is changing anything
- **deps PRs** — a bound change that resolves to something broken

Say the word if you did want the schedule gone and I'll strip it to PR-only.

PR #319, opened 2026-08-19
Port nerve/mcp_server/ to the mcp 2.x lowlevel API

### LINKED ISSUE #316: Fresh install broken by mcp 2.0.0 — ImportError: cannot import name 'request_ctx' from 'mcp.server.lowlevel.server'

alex-clickhouse:
Fixes #316. **Merge first** of the three open PRs — `main` is broken for fresh
installs until this lands.

Now a single commit on `main`. #317 (the `mcp<2` stopgap) is closed, and its commit
has been removed from this branch's history, so the diff is purely the port.

## The bug

A fresh install fails at startup with the `ImportError` in the issue. `mcp` was never
declared in `pyproject.toml` — it arrived transitively via `claude-agent-sdk`, whose
constraint is `mcp<3.0.0,>=1.23.0`. Nothing else in the tree capped it, so once mcp
2.0.0 landed on PyPI (2026-07-28) every fresh resolve picked it up.

Declaring `mcp` explicitly is correct independent of the version question: three
modules under `nerve/mcp_server/` import it directly.

## What changed upstream

mcp 2.0 is an API redesign of the lowlevel server, not a rename:

| | mcp 1.x | mcp 2.x |
|---|---|---|
| Handler registration | `@server.list_tools()` / `@server.call_tool()` decorators | `Server(..., on_list_tools=, on_call_tool=)` constructor callbacks |
| Request context | `request_ctx` contextvar | `ServerRequestContext` passed as the handler's first argument |
| `list_tools` returns | `list[Tool]` | `ListToolsResult` |
| `call_tool` receives | `(name, arguments)` | `CallToolRequestParams` |
| Dispatch table | `server.request_handlers[RequestClass]` | `server.get_request_handler("tools/call")` |
| Handler result | wrapped in the `ServerResult` root union | returned as-is |
| Model attributes | camelCase (`isError`, `clientInfo`) | snake_case; camelCase retained as wire aliases |

`request_ctx` has no replacement — there is no request contextvar anywhere in mcp 2.0
— so `_resolve_client_info` and `_bound_identity_from_request` now take the context
explicitly and `build_ctx_resolver` threads it down from the call handler. Their
bodies are otherwise unchanged.

`StreamableHTTPSessionManager`'s constructor is identical across both majors, so the
transport wiring is untouched.

## Two things worth your attention

**1. mcp 2.x silently drops input validation.** mcp 1.x's `@server.call_tool()`
defaults to `validate_input=True` and runs `jsonschema.validate(arguments,
inputSchema)` before the handler. The 2.x callback does nothing of the kind. Since
nerve used the bare decorator, **every external tool call has been schema-validated
purely by library default** — with no test covering it, precisely because the library
supplied it.

Porting mechanically would have dropped that on an endpoint external MCP clients can
reach, with nothing failing to indicate it. So `build_mcp_server` now validates
explicitly and returns the same `isError` shape as its other failure paths.
`jsonschema` is declared as a direct dependency (it was already an install-time
transitive of claude-agent-sdk) because we now import it. A malformed schema is
treated as our bug and refuses the call rather than skipping validation.

**2. A latent break the existing suite could not see.** The
`clientInfo` → `client_info` rename sits on a line guarded by
`if rctx.session and rctx.session.client_params:`. Every unit test builds a context
with `session=None`, so that line never executed and **all 3303 tests passed with it
broken**. Over real HTTP it raises `AttributeError` inside the resolver, which
`_call_tool` converts into `"Context error"` — meaning every tool call from every
external MCP client would have failed.

What caught it was adding the `tools/call`-over-real-HTTP test that
`test_mcp_http_integration`'s module docstring already claimed to have ("drives a full
`initialize` → `tools/list` → `tools/call` flow") but which was never actually
written. Worth noting as a reviewing lens: unit tests that construct their own context
objects cannot validate context plumbing.

## Testing

- [x] **3310 pass** under mcp 2.0.0 (3303 baseline + 5 validation + 2 HTTP integration), on a fresh `uv pip install -e ".[test]" --refresh` resolve
- [x] `import nerve.gateway.server` clean (the `nerve start` path)
- [x] **Agent-backend bridge**, the cut-over's real risk: nerve serves *every* agent tool through `claude_agent_sdk.create_sdk_mcp_server`, so with `mcp>=2` the SDK's mcp-2.x branch becomes the only path — and it serves the whole Claude backend, not just `/mcp/v1`. Confirmed the SDK reports `MCP_MAJOR == 2` and round-trips a real `tools/list` + `tools/call` against a nerve registry
- [x] **New:** `initialize` → `tools/call` over the real transport, asserting the satellite session id derives from the handshake's `clientInfo` — the observable proof the context arrives populated
- [x] **New:** invalid arguments rejected over real HTTP, handler never entered
- [x] Negative control: forcing `mcp==1.29.0` fails at import (`No module named 'mcp.server.context'`), confirming the `>=2` floor is load-bearing
- [x] Audited for sibling camelCase reads on mcp models; `clientInfo` was the only one. `isError=`/`inputSchema=` appear solely as constructor kwargs, which still work via `populate_by_name`

Frontend untouched (Python only).

## Still worth doing by hand before this leaves draft

Point a real external client (Codex / Claude Code) at `/mcp/v1`. The new integration
test covers the transport with a genuine context, which is most of what I wanted from
a manual run — but a real client exercises protocol negotiation and its own
`clientInfo` shape, and that's the surface where this port's remaining risk lives.

### DIFF
--- a/nerve/mcp_server/http.py
+++ b/nerve/mcp_server/http.py
@@ -29,7 +29,7 @@ import json
 import logging
 from typing import TYPE_CHECKING, Callable
 
-from mcp.server.lowlevel.server import request_ctx
+from mcp.server.context import ServerRequestContext
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 from starlette.types import Receive, Scope, Send
 
@@ -71,22 +71,28 @@ async def _send_status(send: Send, status: int, message: str) -> None:
     await send({"type": "http.response.body", "body": body, "more_body": False})
 
 
-def _resolve_client_info() -> tuple[str | None, str | None, str | None]:
-    """Read client metadata from the active MCP request context.
+def _resolve_client_info(
+    rctx: ServerRequestContext | None,
+) -> tuple[str | None, str | None, str | None]:
+    """Read client metadata from the supplied MCP request context.
 
     Returns ``(client_name, mcp_session_id, request_path)`` — any field
     may be ``None`` if the corresponding data isn't available (e.g.
     during the very first ``initialize`` call the session may not yet
     have ``client_params``).
+
+    mcp 2.x hands the request context to handlers as an argument rather
+    than exposing it through a contextvar, so callers thread it down from
+    the handler. ``None`` is accepted for callers with no live request.
     """
-    try:
-        rctx = request_ctx.get()
-    except LookupError:
+    if rctx is None:
         return None, None, None
 
     client_name: str | None = None
     if rctx.session and rctx.session.client_params:
-        info = rctx.session.client_params.clientInfo
+        # mcp 2.x exposes the model fields as snake_case in Python; the wire
+        # form is still camelCase (`clientInfo`) via the alias generator.
+        info = rctx.session.client_params.client_info
         if info is not None:
             client_name = info.name
 
@@ -105,6 +111,7 @@ def _resolve_client_info() -> tuple[str | None, str | None, str | None]:
 
 def _bound_identity_from_request(
     config: "NerveConfig",
+    rctx: ServerRequestContext | None,
 ) -> tuple[str | None, dict[str, str]]:
     """Session id bound into the request's bearer token, if any.
 
@@ -120,9 +127,7 @@ def _bound_identity_from_request(
     """
     if not config.auth.jwt_secret:
         return None, {}
-    try:
-        rctx = request_ctx.get()
-    except LookupError:
+    if rctx is None:
         return None, {}
     request = getattr(rctx, "request", None)
     if request is None:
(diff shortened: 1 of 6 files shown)


alex-clickhouse:
Retargeted from #317's branch to `main`. Reason: `ci.yml` triggers on
`pull_request: branches: [main]`, so while this was stacked it received **no CI
checks at all** — and of the three PRs in this chain, the port is the one that most
needs them. Stacking was costing exactly the signal it was meant to organise.

Consequence for reviewers: the diff now contains **two commits** — #317's cap
(`0b90148`) and the port itself (`02264c9`). Read `02264c9` alone for the port. If
#317 merges first this collapses to just the port; if it doesn't, this PR delivers
both, which is coherent either way since the port supersedes the cap.

Worth being explicit that #317 is still worth merging on its own: fresh installs are
broken *right now*, that's a one-line change that can go in immediately, and this
port needs real review plus a live-client check first.

alex-clickhouse:
Good catch — fixed in `72f8a1d`, and it's worse than described, so thank you for it.

The floor is now `claude-agent-sdk>=0.2.140`. I checked every release's own `mcp`
constraint to find the right number:

```
0.2.82            mcp>=1.23.0            ← no upper bound at all
0.2.100 … 0.2.139 mcp<2.0.0,>=1.23.0     ← explicitly incompatible
0.2.140           mcp<3.0.0,>=1.23.0     ← first release allowing 2.x
```

So 0.2.82 is the genuinely dangerous one: because it declares no ceiling on `mcp`,
`0.2.82 + mcp 2.x` is a **resolvable** combination rather than one the resolver would
reject. 0.2.100–0.2.139 at least fail loudly.

Confirmed it isn't theoretical — installed that pair and built the in-process server:

```
$ uv pip install 'claude-agent-sdk==0.2.82' 'mcp==2.0.0'    # installs fine
$ python -c "from claude_agent_sdk import create_sdk_mcp_server, tool; ..."
AttributeError: 'Server' object has no attribute 'list_tools'
```

And you're right that the blast radius is bigger than the endpoint this PR ports.
`nerve/agent/backends/claude.py` builds the `"nerve"` MCP server through
`create_sdk_mcp_server` for **every** agent tool, so an SDK without 2.x support takes
down the whole Claude backend, not just `/mcp/v1`. The comment on the floor now records
that reasoning so nobody lowers it back without understanding the coupling.

pufit (approved):

PR #322, opened 2026-08-19
Sidebar: fix scrollbar artifacts, keep sessions with pending work "running"

pufit:
Three sidebar fixes from one report: a stray white dot, a horizontal scrollbar, and sessions dropping out of "Running" the moment their turn ended.

## 1. The white dot in the panel corner

`index.css` styles the scrollbar, its track and its thumb — but never `::-webkit-scrollbar-corner`. Where a horizontal and a vertical scrollbar meet, WebKit painted that square with its own near-white default: a bright dot on a dark panel. Now transparent, like the track.

## 2. The horizontal scrollbar

Not cosmetic — a sizing bug. The `...` load-more row was `w-full` **plus** `mx-1`, i.e. 100% + 8px, so the list overflowed by exactly those 8px (measured live: `clientWidth` 233, `scrollWidth` 241). Sized to `calc(100% - 0.5rem)`, and the list is pinned to `overflow-x-hidden`: it only ever scrolls vertically and rows truncate, so nothing was reachable only by scrolling sideways.

## 3. Sessions with pending work stay "running" — and stay put

A session that scheduled a wakeup, or left a background job in flight, is not idle: it will pick itself back up. Two problems today:

* the moment its turn ended it fell out of the pinned **Running** group, and
* when the continuation landed, the assistant message bumped `updated_at` and yanked the session to the top of the list — reordering the panel for work the user never triggered.

Each row (and each `session_running` event) now carries two live bits:

| field | source | meaning |
|---|---|---|
| `pending_wakeup_at` | `session_wakeups`, `status='pending'` | fire time of the one wakeup a session can have pending |
| `has_background_tasks` | the engine's per-session background-task registry | `run_in_background` Bash/Agent work still live in its CLI |

A row with either is **parked**: it stays in the Running group and shows a **violet** pulse instead of the emerald one, so "working right now" still reads apart from "will resume on its own". Violet was the only free hue — blue is *awaiting your input*, orange *review loop needs you*, emerald *running*, red *error*, yellow *stopped*, accent *unread*. A tooltip names what it is waiting on (`Background job running` / `Scheduled wake-up at 03:24 PM`). Ranked below the error dot on purpose: a failed turn must not be masked by the background work it left behind.

Shipping both bits **with** the transition (not just on the list route) matters because the client deliberately skips the list refetch for the active session.

`updated_at` is now frozen for turns nobody asked for — `internal=True` runs (fired wakeup, restart continuation, star hook) and autonomous drains (a background task settling). The message row is still recorded and `message_count` still advances; only the sidebar's sort key is left alone. A turn the user actually sent bumps it as before.

## Notes

* `GET /api/sessions/search` was duplicating `_decorate` inline; it now calls it, so search rows get the same treatment (including review-loop state) for free.
* Non-active sessions' background-task state is eventually consistent: `background_tasks_update` is broadcast on the session channel only, so a task that settles *without* triggering a continuation turn leaves the dot up until the next list refresh. Making that event global would close it — deliberately out of scope here.

## Testing

* `pytest tests/` — 3351 passed (15 new in `tests/test_parked_sessions.py`: the `updated_at` freeze at the store level and through feed ordering, the engine's pending-work lookup + broadcast payload, and the two fields on every sidebar list route).
* `npm test` — 114 passed (5 new for the client handler that applies the fields).
* `npm run lint` — unchanged (148 pre-existing problems before and after).
* Verified in the browser: the 8px overflow and the corner dot are gone; parked rows render violet inside Running, clearly distinct from the green ones.

### DIFF
--- a/nerve/agent/engine.py
+++ b/nerve/agent/engine.py
@@ -1451,6 +1451,43 @@ class AgentEngine:
         runs to defer the terminal 'done' until spend-relevant work ends."""
         return self._has_live_background_tasks(session_id)
 
+    async def pending_wakeup_at(self, session_id: str) -> str | None:
+        """``fire_at`` of the session's pending wakeup, or ``None``.
+
+        ``ScheduleWakeup`` keeps at most one pending row per session (a new
+        request supersedes the old one), so this is a single timestamp
+        rather than a list. Never raises — a listing/indicator caller must
+        not fail on an enrichment read.
+        """
+        try:
+            rows = await self.db.list_pending_wakeups(session_id)
+        except Exception:  # noqa: BLE001 — enrichment must never break callers
+            logger.debug(
+                "pending wakeup lookup failed for %s", session_id, exc_info=True,
+            )
+            return None
+        return rows[0].get("fire_at") if rows else None
+
+    async def _broadcast_session_running(
+        self, session_id: str, is_running: bool,
+    ) -> None:
+        """Emit the global ``session_running`` event for a turn boundary.
+
+        Carries the session's *pending work* alongside the flag: a turn can
+        end with a wakeup scheduled or a background task still in flight, in
+        which case the session is parked rather than idle and the sidebar
+        keeps showing it as running (in its own colour). Shipping both bits
+        with the transition means the client can redraw the row without a
+        follow-up list fetch — which it skips for the active session.
+        """
+        await broadcaster.broadcast("__global__", {
+            "type": "session_running",
+            "session_id": session_id,
+            "is_running": is_running,
+            "pending_wakeup_at": await self.pending_wakeup_at(session_id),
+            "has_background_tasks": self._has_live_background_tasks(session_id),
+        })
+
     def get_codex_ultracode_run_ids(self, session_id: str) -> set[str] | None:
         """Ultracode run ids the session's live Codex client attributed to
         its current turn, or None when unavailable (no client / claude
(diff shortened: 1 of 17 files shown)


nerve/agent/engine.py
```diff
@@ -2816,7 +2854,13 @@ async def _run_inner(
         # run_in_background task settles, the CLI runs an autonomous turn
         # which the idle stream watcher drains to the UI — no Nerve-side
         # output-file polling needed (the old regex watcher lived here).
-        await self._finalize_turn(session_id, st, channel)
+        #
+        # ``internal`` turns carry a synthetic trigger nobody typed (a fired
+        # wakeup, a restart continuation, a star hook): they resume work the
+        # session was already doing, so they must not re-sort the sidebar.
+        await self._finalize_turn(
+            session_id, st, channel, bump_updated_at=not internal,
```
alex-clickhouse: I think it would be better to add `bump_updated_at` as a param to run() instead of overloading `internal` here. A couple of things that won't be bumped, when I think they should:

* Starring a session, which is marked internal: https://github.com/ClickHouse/nerve/blob/a587b067737f70d4532dc90fcf7ec68ae5c02792/nerve/gateway/routes/sessions.py#L483
* Waking up a run-later task: https://github.com/ClickHouse/nerve/blob/a587b067737f70d4532dc90fcf7ec68ae5c02792/nerve/cron/service.py#L1352

Both of those are user-requested actions that should bump `updated_at`.

Alternatively, we could just explicitly bump `updated_at` at both of those sites and leave the `internal` check as it is?

alex-clickhouse (approved):

PR #408, opened 2026-08-24
feat(install): unattended installation mode

nikita-vanyasin:
## Problem

`install.sh` cannot complete without a terminal, so it can't be used from cloud-init, Ansible, a Dockerfile or an image build. Two separate things block it.

**1. The tty reclaim is fatal without a controlling terminal.**

```bash
if [ ! -t 0 ] && [ -e /dev/tty ]; then
    exec < /dev/tty
fi
```

`/dev/tty` exists on any normal system, but a process started by a daemon has no controlling terminal, so the open fails with ENXIO. Under `set -e` that ends the install before the first dependency check. `-e /dev/tty` tests that the device node is there, which is not the same question.

**2. There is no way to skip the wizard.**

Past the dependency phase `run_init` always runs the interactive `nerve init`. `NERVE_YES=1` does not help — it makes `confirm "Run the setup wizard now?"` return true, so the wizard runs with no one to answer it and the install hangs. `offer_start` then does the same for "Start Nerve now?".

`nerve init --non-interactive` already exists and reads everything from the environment; the installer just had no way to reach it.

## Change

Adds `--non-interactive` / `NERVE_NON_INTERACTIVE=1`:

- implies `NERVE_YES=1` and skips the tty reclaim entirely
- `run_init` calls `nerve init --non-interactive` and fails loudly rather than leaving a half-configured install
- `offer_start` leaves the daemon stopped, since these installs are normally followed by a service manager that owns the process. `NERVE_START=1` starts it from the installer instead.
- an existing install still takes the upgrade path and keeps its configuration

The tty reclaim is now probed in a subshell before it is used, so an interactive run behaves as before and a run without a terminal continues instead of dying.

```bash
curl -fsSL https://raw.githubusercontent.com/ClickHouse/nerve/main/install.sh \
  | NERVE_NON_INTERACTIVE=1 \
    NERVE_INSTALL_DIR=/home/agent/nerve \
    NERVE_MODE=worker \
    NERVE_PROVIDER=bedrock \
    NERVE_AWS_REGION=eu-central-1 \
    bash
```

`docs/setup.md` gets an "Unattended installation" section listing the variables the wizard would otherwise ask for.

## Notes

- No behaviour change for an interactive install: `NON_INTERACTIVE` defaults to 0 and every new branch is behind it.
- `NERVE_PASSWORD` is read once by `nerve init` and stored only as a bcrypt hash, so an unattended caller can pass it in and drop it immediately. Called out in the docs.

## Testing

- `bash -n install.sh`
- Verified `NERVE_NON_INTERACTIVE=1` sets `AUTO_YES=1` and that `confirm` then returns without reading stdin.
- Verified the subshell probe swallows an unopenable path and lets the script continue, where a bare `exec` redirection does not.
- Not yet run as a full unattended install on a clean host.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,9 +7,13 @@
 #   curl -fsSL https://raw.githubusercontent.com/ClickHouse/nerve/main/install.sh | bash
 #
 # Environment variables:
-#   NERVE_INSTALL_DIR  — Where to clone the repo (default: ~/nerve)
-#   NERVE_BRANCH       — Git branch to install (default: main)
-#   NERVE_YES          — Set to 1 to skip all confirmations
+#   NERVE_INSTALL_DIR      — Where to clone the repo (default: ~/nerve)
+#   NERVE_BRANCH           — Git branch to install (default: main)
+#   NERVE_YES              — Set to 1 to skip all confirmations
+#   NERVE_NON_INTERACTIVE  — Set to 1 for unattended install: implies NERVE_YES,
+#                            configures from env (see docs/setup.md), never prompts
+#   NERVE_START            — Set to 1 to start the daemon after an unattended
+#                            install; leave unset when a service manager owns it
 #
 set -euo pipefail
 
@@ -24,7 +28,9 @@ MIN_PYTHON_MINOR=13
 PREFERRED_PYTHON_MINOR=13
 # Vite 7 (see web/package.json) requires Node 20.19+ or 22.12+.
 MIN_NODE_VERSION="20.19.0"
+NON_INTERACTIVE="${NERVE_NON_INTERACTIVE:-0}"
 AUTO_YES="${NERVE_YES:-0}"
+[ "$NON_INTERACTIVE" = "1" ] && AUTO_YES=1
 IS_UPGRADE=0
 
 # --- Colors ---
@@ -499,6 +505,15 @@ run_init() {
         return
     fi
 
+    if [ "$NON_INTERACTIVE" = "1" ]; then
+        step "Running Nerve setup (non-interactive)"
+        cd "$INSTALL_DIR" || exit 1
+        # Fail loudly: a half-configured unattended install is worse than none.
+        "$nerve_bin" -c "$INSTALL_DIR" init --non-interactive
+        INIT_COMPLETED=1
+        return
+    fi
+
     # Clear boundary between installation and configuration: everything is
     # installed at this point — the wizard is optional and resumable.
     printf "\n"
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -24,7 +28,9 @@ MIN_PYTHON_MINOR=13
 PREFERRED_PYTHON_MINOR=13
 # Vite 7 (see web/package.json) requires Node 20.19+ or 22.12+.
 MIN_NODE_VERSION="20.19.0"
+NON_INTERACTIVE="${NERVE_NON_INTERACTIVE:-0}"
```
alex-clickhouse, line 34: Could we make dependency elevation honor unattended mode too? The official `ubuntu:24.04` image runs as root and has `apt-get`, but has neither `sudo` nor `git`, so the current `HAS_SUDO=0` guard aborts even though it can install packages directly. In a container with `sudo` added, `sudo apt-get -y` also allowed `tzdata` to emit its `Geographic area:` debconf prompt because `DEBIAN_FRONTEND` was stripped; `-y` only answers apt's own confirmation.

One way to centralize both concerns:

```bash
run_as_root() {
    if [ "$(id -u)" -eq 0 ]; then
        "$@"
    elif ! command_exists sudo; then
        error "Root privileges are required, but sudo is not available."
        return 1
    elif [ "$NON_INTERACTIVE" = "1" ]; then
        # Fail instead of waiting for a sudo password in automation.
        sudo -n "$@"
    else
        sudo "$@"
    fi
}

run_debian_command() {
    if [ "$NON_INTERACTIVE" = "1" ]; then
        # Set this after the sudo boundary so sudo cannot strip it.
        run_as_root env DEBIAN_FRONTEND=noninteractive "$@"
    else
        run_as_root "$@"
    fi
}
```

The apt paths can then use:

```bash
run_debian_command apt-get update -qq
run_debian_command apt-get install -y -qq git

curl -fsSL https://deb.nodesource.com/setup_lts.x \
    | run_debian_command bash -
run_debian_command apt-get install -y -qq nodejs
```

The Python apt path should use the same wrapper, while `dnf`/`pacman`/`zypper` can use `run_as_root`. This keeps interactive sudo behavior unchanged, lets root images install directly, and guarantees unattended sudo never prompts.

install.sh
```diff
@@ -499,6 +505,15 @@ run_init() {
         return
     fi
 
+    if [ "$NON_INTERACTIVE" = "1" ]; then
+        step "Running Nerve setup (non-interactive)"
+        cd "$INSTALL_DIR" || exit 1
+        # Fail loudly: a half-configured unattended install is worse than none.
+        "$nerve_bin" -c "$INSTALL_DIR" init --non-interactive
```
alex-clickhouse: Could we resolve the config directory once, for example `CONFIG_DIR="${NERVE_CONFIG_DIR:-$INSTALL_DIR}"`, and use it consistently? This explicit `-c "$INSTALL_DIR"` ignores `NERVE_CONFIG_DIR` during init, but `offer_start` later calls `nerve start` without `-c`, where the same environment variable wins and may select a different config. The upgrade check also only looks at `$INSTALL_DIR/config.local.yaml`.

I do not think the config directory needs to be mandatory—the install directory is a safe default—but upgrade detection, init, start, and the summary should all use the same resolved value.

install.sh
nikita-vanyasin: Done

`run_as_root` runs directly when `id -u` is 0, errors when there is no sudo and no root, and uses `sudo -n` under `--non-interactive` so a missing password rule fails immediately rather than waiting. `run_debian_command` wraps it with `DEBIAN_FRONTEND=noninteractive` past the sudo boundary.

Every apt path now goes through `run_debian_command`.

install.sh
nikita-vanyasin: Fixed. `CONFIG_DIR="${NERVE_CONFIG_DIR:-$INSTALL_DIR}"` is resolved once next to `INSTALL_DIR`.

alex-clickhouse (approved):

alex-clickhouse:
LGTM, thank you for the PR!

PR #410, opened 2026-08-25
sources: preserve cursor when inbox persistence fails

fallintoplace:
## What changed

- **Fail closed** when inbox persistence fails.
- Let the database transaction roll back the full batch.
- Return an ingestion error and keep the source cursor unchanged.
- Retry the batch on the next healthy run. Existing dedup keeps retries safe.

## Why

A bad row used to be logged and skipped. The rest of the batch still committed and the cursor moved forward, so the bad row could be lost permanently.

## Tests

- `uv run pytest tests/test_source_runner_persistence.py tests/test_source_filters.py tests/test_source_resurface.py tests/test_db_atomicity.py`
- `uv run python -m compileall -q nerve/db/sources.py nerve/sources/runner.py tests/test_source_runner_persistence.py`

### DIFF
--- a/nerve/db/sources.py
+++ b/nerve/db/sources.py
@@ -3,7 +3,10 @@
 from __future__ import annotations
 
 import json
+import logging
 from datetime import datetime, timedelta, timezone
+
+logger = logging.getLogger(__name__)
 
 
 class SourceStore:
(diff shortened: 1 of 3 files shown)


alex-clickhouse (approved):

alex-clickhouse:
LGTM, thank you!

PR #415, opened 2026-08-27
Telegram: route a reply to the session that sent the replied-to message

serxa:
## What

Adds **opt-in** reply routing for Telegram, behind a new setting `telegram.reply_routes_to_origin_session` (**default `false`**).

- **Off (default):** unchanged behavior — a reply is *context-only*: it quotes the replied-to message (and pulls in its attachments) and lands in the chat's **active** session, exactly like a plain message. This is the existing/intended behavior, so nothing changes for anyone who doesn't opt in.
- **On:** a **reply** routes to the **session that produced the replied-to message** (and makes it active), so you can answer cron/background sessions, or reply to your own earlier message to jump back into whatever session it went to — without an explicit `/session` switch.

## How

When the flag is off, none of the machinery below runs and inbound replies are delivered to the active session (`session_id` stays unset). When on:

- **Recording (both directions):** a bounded LRU keyed by `(chat_id, message_id)` records the session for every message the bot sends (`TelegramChannel.send` + `NotificationService._deliver_telegram`, the cron case) and for every inbound message's landing session (`_delivery_session`). The key is `(chat_id, message_id)` because Telegram message ids are unique only within a chat.
- **Inbound:** `_resolve_reply()` classifies a message three ways — **not a reply** → active session; **reply to a mapped, live session** → route there + make active; **reply that can't be resolved** → reject.
- The flag is read live per message, so toggling it takes effect on the next message (no restart).

## Safety (when on): an unresolvable reply is rejected, never mis-routed

A reply we can't tie to a live session — the target isn't in the map (sent before the daemon started, since the map is in-memory; or evicted from the LRU), its session is archived/gone, or the chat id doesn't match — is **not** delivered to the active session (that would be a silent mis-route). Instead the bot sends a static, **no-LLM** nudge ("couldn't find the session… use `/sessions` and resend") and drops the message without starting a turn.

## Behavior choice (when on)

Routing is **sticky** — a reply also makes the target session active (like `/session <id>`), so a following plain message continues there.

## Config

| Setting | Type | Default | Meaning |
|---|---|---|---|
| `telegram.reply_routes_to_origin_session` | bool | `false` | Off = reply is context-only, lands in active session. On = reply routes to the replied-to message's session (sticky); unresolvable reply is refused. |

## Testing

- `tests/test_telegram_reply_routing.py` (20 tests): record/lookup, `(chat_id, message_id)` cross-chat isolation, LRU eviction/recency, three-way `_resolve_reply`, `_delivery_session`, reply-to-own-message routing, `send()` recording, and the flag gate (enabled reflects config; recording skipped when off).
- Full suite: **3378 passed**; the only 2 failures (`test_config_env`, `test_lockdown`) reproduce on clean `main` in this environment and are unrelated to this change.

### DIFF
--- a/nerve/channels/telegram.py
+++ b/nerve/channels/telegram.py
@@ -491,12 +491,13 @@ class TelegramChannel(BaseChannel):
             collections.OrderedDict()
         )
         self._message_cache_max = 200
-        # Reply routing: telegram message_id -> (chat_id, session_id). Records
-        # both messages the bot sent and the session each inbound message landed
-        # in, so a user REPLY targets that session — a cron/background session's
+        # Reply routing: (chat_id, message_id) -> session_id. Records both
+        # messages the bot sent and the session each inbound message landed in,
+        # so a user REPLY targets that session — a cron/background session's
         # message, or the user's own earlier message — instead of the chat's
-        # active session. Bounded LRU.
-        self._reply_routes: collections.OrderedDict[int, tuple[int, str]] = (
+        # active session. Keyed by (chat_id, message_id) because Telegram
+        # message ids are unique only within a chat, not across chats. Bounded LRU.
+        self._reply_routes: collections.OrderedDict[tuple[int, int], str] = (
             collections.OrderedDict()
         )
         self._reply_routes_max = 2000
@@ -994,13 +995,16 @@ class TelegramChannel(BaseChannel):
         """Remember which session produced an outbound Telegram message.
 
         Lets a later user REPLY to that message route back to the
-        originating session (see ``_resolve_reply_session``). No-op when the
-        session is unknown. Bounded LRU keyed by ``message_id``.
+        originating session (see ``_resolve_reply``). No-op when the session
+        is unknown. Bounded LRU keyed by ``(chat_id, message_id)`` — message
+        ids repeat across chats, so keying on the id alone would let one chat
+        clobber another chat's identically-numbered message.
         """
         if not session_id:
             return
-        self._reply_routes[message_id] = (chat_id, session_id)
-        self._reply_routes.move_to_end(message_id)
+        key = (chat_id, message_id)
+        self._reply_routes[key] = session_id
+        self._reply_routes.move_to_end(key)
         while len(self._reply_routes) > self._reply_routes_max:
             self._reply_routes.popitem(last=False)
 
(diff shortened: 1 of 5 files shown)


nerve/channels/telegram.py
```diff
@@ -966,6 +984,85 @@ def _cache_message(self, message_id: int, chat_id: int, text: str) -> None:
...
+    ) -> None:
+        """Remember which session produced an outbound Telegram message.
+
+        Lets a later user REPLY to that message route back to the
+        originating session (see ``_resolve_reply_session``). No-op when the
+        session is unknown. Bounded LRU keyed by ``message_id``.
+        """
+        if not session_id:
+            return
+        self._reply_routes[message_id] = (chat_id, session_id)
```
alex-clickhouse: I think the key here needs to be `(chat_id, message_id)`, as message_ids are not unique by themselves.

alex-clickhouse:
>Routing is sticky — a reply also makes the target session active (like /session <id>), so a following plain message continues there.

Is this actually desirable? I think it may be quite surprising/unexpected behavior for users. Open to being convinced otherwise though.

serxa:
I was initially thinking that sticky is bad as well, but TBH, if it's not sticky, then you have to keep two things in your head - where you are answering and where the current active session is. This is tricky if you have many sessions. Imagine you started something and then replied to 5 different sessions. At this point, you probably do not remember where you started. Also, when you later look at chat history, it is completely unclear where the message goes.

On the other hand, sticky resolves everything. This is the natural way chat works. Wherever you are, just continue the conversation in a way consistent with your previous message. It's 100% better IMO and keeps chat history clear and understandable.

pufit:
I personally would like to have this feature. But I talked with Nerve contributor @constkolesnyak before who originally introduced most of the Telegram integration and he said that the current behavior is intended. 

Let's put it under a feature flag in settings?

serxa:
> current behavior is intended.

Could you please elaborate on it a bit? If I understood correctly, currently replies allow to enrich somehow what is routed to the current session. I didn't use it, but anyway, this is unchanged, I believe. You can reply to a message generated on the same session - this is exactly the previous behavior.

Slop mode:

What you lose by enabling unconditionally:

Quoting an arbitrary / cross-session message into your current session. A reply now either switches you to that message's session, or — if it's unmapped — rejects. So you can't reply to something from another session (or an old one) just to pull it in here. Workaround = copy-paste or a plain message.

The "a reply always delivers" guarantee. An unmapped reply now rejects instead of landing in active. Sharpest edge: the map is in-memory, so right after any restart, replying to any pre-restart message (including the bot's last one) rejects until you exchange a few fresh messages — same for your own old messages or bot messages past the 2000 LRU cap.

serxa:
> Let's put it under a feature flag in settings?

I'm okay with this if that's what we want.

alex-clickhouse:
Can we make the flag purely additive? When it is on, route the reply if we can resolve it, and otherwise deliver it the way we do today into the active session. Then both of your points hold in either mode:

- Quoting an arbitrary or cross-session message still works. It lands in the active session, as before.
- A reply always delivers. Replying to a pre-restart message, or to one past the LRU cap, is not refused.

That also drops the reject path and the nudge message.

serxa:
No, that doesn't work. In such a way we will break both old and new behaviour. Reply will either go to active session or switch it into another session. And the decision will depend on thing you cannot predict: if nerve is aware of the session that produced the message you reply to.

Also, currently we store that info in memory and it dies with nerve restart. But even if we fix it and we will remember source of all messages - this will be equal to removing old behaviours completely.

I think approaches are alternative.

PR #420, opened 2026-09-02
memory: allow a self-hosted embeddings endpoint

nikita-vanyasin:
## Problem

The memU bridge hardcoded the embeddings host, so embeddings only ever worked against OpenAI:

```python
if self.config.openai_api_key:
    llm_profiles["embedding"] = {
        "base_url": "https://api.openai.com/v1",
        ...
```

On a Bedrock install that means no vector recall at all. `_BedrockLLMClient.embed()` raises `NotImplementedError` and the client is `AsyncAnthropicBedrock`, which only speaks Anthropic models, so Titan is not reachable through it either. Those installs fall back to LLM-based recall, which spends Sonnet or Haiku tokens on every query. Any other install has to send memory text to OpenAI and hold a static third-party key.

memU needs no change for this. `OpenAISDKClient.__init__` already takes a `base_url` and builds a normal `AsyncOpenAI`, then embeds with `client.embeddings.create(model=self.embed_model, input=inputs)`. Nerve was the only thing pinning the host.

## Change

New `memory.embed_base_url` accepts any endpoint implementing OpenAI `/v1/embeddings`, self-hosted or proxied.

The gate widens from `openai_api_key` to a key *or* a base URL, so an endpoint on its own enables embeddings. It does its own auth, which makes `openai_api_key` optional; `AsyncOpenAI` still rejects an unset key, hence the `"placeholder"`.

`run_non_interactive` reads `NERVE_EMBEDDINGS_API_ENDPOINT` and `NERVE_EMBEDDINGS_MODEL`. The model variable is needed because a self-hosted gateway will not serve `text-embedding-3-small` under that name, and without it the hardcoded default is baked in.

The profile logic moved into a pure `build_embedding_profile()`. It was previously inline in `_initialize_impl`, which stands up a real `MemoryService` and so could not be tested.

Two follow-on fixes:

- `_check_openai_key` probes the configured host rather than always OpenAI. It would otherwise report a failure at the end of `nerve init` for any custom endpoint.
- `nerve doctor` reports the endpoint, and returns `[ERR]` when a base URL is set with no `embed_model`.

## Compatibility

Defaults are unchanged: no base URL means `api.openai.com` gated on the key, as before. `embed_base_url` is a new dataclass field, so unknown-key validation picks it up automatically and older configs parse identically.

An endpoint set without `embed_model` disables embeddings with a warning rather than calling the API with `model=""`, which fails opaquely at query time. `from_dict` coerces `None` to `""`, so a bare `embed_base_url:` key does not become the string `"None"`.

`embed_base_url` is written to the portable `settings.yaml` layer, since it describes the deployment. The Bedrock model rewrite only touches `recall_model`, `memorize_model` and `fast_model`, so it leaves this key alone.

## Tests

13 new tests. `TestBuildEmbeddingProfile` covers the gating matrix: endpoint alone, key alone, endpoint overriding a set key, and both no-model paths. `TestNonInteractiveEmbeddingsEndpoint` covers the env plumbing, including that the Bedrock rewrite leaves the embed keys intact and that an omitted variable leaves the config key absent. Config parsing covers defaults, whitespace and `None`.

Full suite: 3378 passed.

Three `TestCredentialWaterfall` tests fail on my machine, but they also fail on a clean `main` worktree and still fail with `ANTHROPIC_API_KEY` unset, so a locally stored credential is leaking into `patch.dict(..., clear=False)`. Unrelated to this change and expected to pass in CI.

## Untested assumption

This assumes the target endpoint returns the OpenAI response shape, `data[].embedding`. A gateway that differs needs its own client. Bedrock embeddings through a separate `bedrock-runtime` call against `amazon.titan-embed-text-v2:0` would be the fallback for that case, and is out of scope here.

### DIFF
--- a/nerve/memory/memu_bridge.py
+++ b/nerve/memory/memu_bridge.py
@@ -87,10 +87,40 @@ def _is_sqlite_locked_error(exc: BaseException) -> bool:
 # Semantic dedup threshold — set from config at init time.
 _SEMANTIC_DEDUP_THRESHOLD = 0.85
 
+# Fallback embeddings endpoint when memory.embed_base_url is unset.
+_OPENAI_EMBED_BASE_URL = "https://api.openai.com/v1"
+
 # Max characters for a category breadcrumb in recall() output. A category
 # summary is a whole rolled-up topic document (often 5–20KB); recall must
 # surface it as a short navigable pointer, not dump the document.
 _CATEGORY_BREADCRUMB_MAXLEN = 200
+
+
+def build_embedding_profile(
+    *, openai_api_key: str, embed_base_url: str, embed_model: str,
+) -> dict[str, str] | None:
+    """memU ``embedding`` LLM profile, or None for LLM-based recall.
+
+    Requires ``embed_model`` plus either a key or a base URL. Any endpoint
+    implementing OpenAI ``/v1/embeddings`` is accepted.
+    """
+    if not (openai_api_key or embed_base_url):
+        return None
+    base_url = embed_base_url or _OPENAI_EMBED_BASE_URL
+    if not embed_model:
+        # Without this the API is called with model="".
+        logger.warning(
+            "Embeddings endpoint %s configured but memory.embed_model is empty "
+            "— embeddings disabled, falling back to LLM-based recall", base_url,
+        )
+        return None
+    return {
+        "base_url": base_url,
+        # AsyncOpenAI rejects an unset key; the endpoint may ignore it.
+        "api_key": openai_api_key or "placeholder",
+        "embed_model": embed_model,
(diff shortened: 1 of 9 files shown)


nerve/memory/memu_bridge.py
alex-clickhouse, line 1753: This still uses the old gate based on the openai_api_key presence.

alex-clickhouse, line 1781: Same as above.

alex-clickhouse, line 1847: Same as above.

alex-clickhouse, line 3529: Same as above.

PR #441, opened 2026-09-13
Local multi-user 6/6: First-account claim

alex-clickhouse:
Part 6 of 6 of the local multi-user stack. This ships one browser setup responsibility only: securely claiming the first account of a passwordless installation.

## What

- Generate and persist a setup token while the sole account is unclaimed; expose it through local CLI status, never URLs, API responses, logs, or browser persistence.
- Require a non-empty setup token in every claim request, regardless of peer address or headers. Validate against the persisted value with a constant-time comparison and fail closed when none is stored.
- `POST /api/setup/claim` atomically names and secures the existing account, optionally sets its display name, deletes the setup secret with secure-delete semantics in the same transaction, and returns only a new client session token.
- Guarantee exactly one winner for concurrent claims and roll the whole claim back if token invalidation fails.
- Add a per-account session epoch in `v049_session_epoch.py`; typed and legacy pre-claim sessions stop working after claim while upgrades remain at epoch zero.
- Guard account mutations inside their write transactions against claim races.
- Register WebSockets before the last admission check, re-check credential epoch/enablement before every frame, proactively close revoked sockets after claim, and retain the per-frame guard if proactive closure fails.
- Route `login === "none"` to a small claim form. After success the browser discards form token state, stores only the client session, refreshes canonical auth through `/api/accounts/me`, refreshes actors, and routes to chat.
- Keep public draft shapes exact: auth status has two fields, accounts seven, and actor refs three.

## Security model

The setup token is the sole claim authority. Origin, Host, forwarding, DNS, loopback, and proxy classification are intentionally irrelevant; a hostile-header test pins that property. Remote setup must use HTTPS or a protected tunnel because the token is a bearer credential.

## Screenshots

Captured against a throwaway instance — its own `NERVE_HOME`, config directory and workspace, on a loopback port — in a synthetic fresh-install state. No setup token, API key, real name or internal address appears in it. Pinned to [`92745e4`](https://github.com/ClickHouse/nerve/tree/92745e4359a720def0530b8ae177f1134294b1b5/screenshots) on `alex/mu-pr-screenshots`, an images-only branch that is not meant to be merged; see its README for provenance.

**A fresh passwordless install**, recaptured from this PR's tip, `2f8c72e`. `/setup` renders the claim form alone: username, password, optional display name, and the **mandatory** setup token — required here however the page was reached, with no locality or origin exemption. The token field is empty, no credential is shown, and the submit stays disabled until it is filled in. Every other write is withheld until the claim, because the claim is the only one anybody may perform.

The claim form on an unclaimed instance, with an empty and required setup-token field

## Verification

- Focused backend claim/account/socket/startup suite: 100 passed.
- Focused frontend setup/auth/actor suite: 48 passed.
- Final backend: `pytest tests/ -v` — 3,812 passed, 1 pre-existing warning.
- Final frontend: `npm run test` — 345 passed.
- `npm run lint`: unchanged baseline of 148 diagnostics (131 errors, 17 warnings).
- `npm run build`: passed.
- Independent backend/security audit found no remaining defect after the restack integration fixes.

## Reviewer notes

- Migration numbers `v047`-`v049` are placeholders pending merge order against the open Slack stack.
- This PR deliberately does not provide permanent browser administration or a browser-triggered daemon restart.

## Stack

1. #436 — Accounts and actor identity
2. #437 — Actor on the request
3. #438 — Account management
4. #439 — Attribution persistence
5. #440 — Attribution display
6. #441 — First-account claim ← this PR


docs/setup.md
```diff
@@ -220,6 +221,49 @@ nerve start              # Start the server
 # Open http://localhost:8900
 ```
 
+## Claiming an instance from a browser
```
alex-clickhouse: Tighten up the prose in this section

